PR #1, opened 2017-05-02
Fix initialization error for number format 40.

ZacharyPF:
Fix load error for the default number format #40 [#,##0.00;[Red](#,##0.00)]


EPPlus/Style/XmlAccess/ExcelNumberFormatXml.cs
```diff
@@ -13,17 +13,17 @@
```
emdelaney: It's not essential, but while you're in here it would be nice to add some regions / bring this file into compliance with the coding standard.

PR #2, opened 2017-05-05
Fix an issue where deleting a table would not actually delete the table from the underlying xml.

emdelaney:
ExcelTableCollection.Delete() is the only method in that file with code changes (The rest is whitespace conversion).

### DIFF
--- a/EPPlus/Table/ExcelTableCollection.cs
+++ b/EPPlus/Table/ExcelTableCollection.cs
@@ -37,212 +37,217 @@ using System.Xml;
 
 namespace OfficeOpenXml.Table
 {
-    /// <summary>
-    /// A collection of table objects
-    /// </summary>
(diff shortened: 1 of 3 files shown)


EPPlus/Table/ExcelTableCollection.cs
```diff
@@ -37,212 +37,217 @@
 
 namespace OfficeOpenXml.Table
```
ZacharyPF: This file is why I don't like to reformat in the middle of changes. I have no idea what your actual changes are in here :p

emdelaney: Yeah, I absentmindedly auto-formatted at some point while working in it. VS was smart enough to ignore the whitespace changes, but Github wasn't.

PR #5, opened 2017-05-10
Fixed issue where date strings were not compared correctly when using…

maxackley:
… logic functions like OR and IF.

Fixed issue where Excel format strings with TEXT function were not properly formatting dates.

### DIFF
--- a/EPPlus/ExcelRangeBase.cs
+++ b/EPPlus/ExcelRangeBase.cs
@@ -1019,7 +1019,7 @@ namespace OfficeOpenXml
 			return ExcelRangeBase.FormatValue(v, nf, forWidthCalc);
 		}
 
-		private static string FormatNumericValue(object value, ExcelFormatTranslator nf, string format)
+		private static bool TryFormatNumericValue(object value, ExcelFormatTranslator nf, string format, out string formatted)
 		{
 			double d;
 			try
@@ -1028,20 +1028,25 @@ namespace OfficeOpenXml
 			}
 			catch
 			{
-				return string.Empty;
+				formatted = string.Empty;
+				return false;
 			}
 			if (nf.DataType == eFormatType.Number)
 			{
 				if (string.IsNullOrEmpty(nf.FractionFormat))
-					return d.ToString(format, nf.Culture);
+					formatted = d.ToString(format, nf.Culture);
 				else
-					return nf.FormatFraction(d);
+					formatted =  nf.FormatFraction(d);
+				return true;
 			}
-			else
+			else if (nf.DataType == eFormatType.DateTime)
 			{
 				var date = DateTime.FromOADate(d);
-				return date.ToString(format, nf.Culture);
+				formatted = date.ToString(format, nf.Culture);
+				return true;
 			}
+			formatted = string.Empty;
+			return false;
 		}
 
 		private static string FormatDateValue(DateTime date, ExcelFormatTranslator nf, string format)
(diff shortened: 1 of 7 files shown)


EPPlus/ExcelRangeBase.cs
```diff
@@ -1014,104 +1016,78 @@ private string GetFormattedText(bool forWidthCalc)
...
-				format = nf.NetFormatForWidth;
-				textFormat = nf.NetTextFormatForWidth;
+				d = Convert.ToDouble(value);
 			}
-			else
+			catch
 			{
-				format = nf.NetFormat;
-				textFormat = nf.NetTextFormat;
+				return string.Empty;
```
emdelaney: Should we be letting the error bubble up here? It seems kind of strange to me, in the abstract, to return an empty string as a sort of error case. This might be a candidate for a TryFormat...() pattern method with an out parameter.

EPPlus/ExcelRangeBase.cs
```diff
@@ -1014,104 +1016,78 @@ private string GetFormattedText(bool forWidthCalc)
...
+			if (value is DateTime)
+				return ExcelRangeBase.FormatDateValue((DateTime)value, nf, format);
+			else if (ConvertUtil.TryParseDateString(value, out DateTime date))
+				return ExcelRangeBase.FormatDateValue(date, nf, format);
+			else if (value is TimeSpan)
+				return ExcelRangeBase.FormatTimeSpanValue((TimeSpan)value, nf, format);
+			else if (ConvertUtil.IsNumeric(value) && (nf.DataType == eFormatType.Number || nf.DataType == eFormatType.DateTime))
+				return ExcelRangeBase.FormatNumericValue(value, nf, format);
+			else
+				return textFormat == string.Empty ? value.ToString() : string.Format(textFormat, value);
```
emdelaney: Can textFormat ever be null?

EPPlusTest/ExcelRangeBaseTest.cs
```diff
@@ -452,5 +453,73 @@ public void AutoFitHandlesRotatedContents()
 			}
 		}
 		#endregion
+
+		#region FormatValue Tests
+		[TestMethod]
+		public void FormatValueDateTime()
```
emdelaney: Do we need to add a test for a culture that uses a different number format, such as German?

EPPlusTest/Utils/ConvertUtilTest.cs
```diff
@@ -73,5 +73,48 @@ public void TryParseDateString()
...
+			// Excel treats dates  and date strings as numeric.
+			Assert.IsTrue(ConvertUtil.IsNumeric("1/1/2000"));
+			Assert.IsTrue(ConvertUtil.IsNumeric(new DateTime(2000, 1, 1)));
+			Assert.IsTrue(ConvertUtil.IsNumeric(new TimeSpan(5, 0, 0, 0)));
+		}
+
+		[TestMethod]
+		public void GetValueDouble()
+		{
+			Assert.IsTrue(((int)3).GetType().IsPrimitive);
```
emdelaney: Are these IsPrimitive assertions necessary? They seem to be testing intrinsic C# functionality.

EPPlus/Utils/ConvertUtil.cs
```diff
@@ -64,49 +70,42 @@ internal static bool TryParseDateString(object candidate, out DateTime result)
...
+		/// Convert an object value to a double.
 		/// </summary>
-		/// <param name="v"></param>
-		/// <param name="ignoreBool"></param>
-		/// <param name="retNaN">Return NaN if invalid double otherwise 0</param>
-		/// <returns></returns>
-		internal static double GetValueDouble(object v, bool ignoreBool = false, bool retNaN = false)
+		/// <param name="value">The value to convert to a double.</param>
+		/// <param name="ignoreBool">If the value is a boolean, indicates the boolean value should be ignored and 0 returned instead.</param>
+		/// <param name="returnNaN">If true, returns NaN if the double is invalid; otherwise returns 0.</param>
```
emdelaney: Otherwise, this returns zero if the double is invalid.

EPPlus/Utils/ConvertUtil.cs
```diff
@@ -64,49 +70,42 @@ internal static bool TryParseDateString(object candidate, out DateTime result)
...
-					else if (v is TimeSpan)
-					{
-						d = DateTime.FromOADate(0).Add((TimeSpan)v).ToOADate();
-					}
+					if (value is DateTime)
+						result = ((DateTime)value).ToOADate();
+					else if (ConvertUtil.TryParseDateString(value, out DateTime date))
+						result = date.ToOADate();
+					else if (value is TimeSpan)
+						result = DateTime.FromOADate(0).Add((TimeSpan)value).ToOADate();
```
emdelaney: Consider using the C#7 type casting paradigm here: `else if (value as TimeSpan duration)` (see fig. 5: https://msdn.microsoft.com/en-us/magazine/mt790184.aspx ).

EPPlusTest/Utils/ConvertUtilTest.cs
maxackley: Nope, these were accidental leftovers from testing.

PR #7, opened 2017-05-16
Refactor the tests in DateTimeFunctionTests.cs into classes for each Excel function being tested.

emdelaney:


emdelaney:
There were no code changes here; the only changes involve moving code around.

PR #8, opened 2017-05-17
Bugs/691 days360 optional arguments

EvanSchallerer:
- [x] Adds support for optional first parameters
- [x] Adds Days360 tests with optional parameters
- [x] Fixes test class that weren't running due to missing annotations and public access modifiers


EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/Days360Tests.cs
```diff
@@ -102,6 +103,28 @@ public void Days360ShouldHandleFebWithUsMethodSpecified2()
 		}
 
 		[TestMethod]
+		public void Days360ShouldHandleNullFirstDateArgument()
```
emdelaney: Do we need to add any tests for non-end-of-month days to feel good about this function?

EPPlus/FormulaParsing/Excel/Functions/DateTime/Days360.cs
```diff
@@ -16,11 +16,8 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
 		{
 			if (ValidateArguments(arguments, 2) == false)
```
emdelaney: While you're in here, if you could tidy this file up to meet our coding standard (this.ValidateArguments, this.CreateResult, #regions as necessary, etc) that would be awesome :)

PR #10, opened 2017-05-18
Bugs/710 improve tests for weekday function

colewfrench:


EPPlus/FormulaParsing/Excel/Functions/DateTime/Weekday.cs
```diff
@@ -29,21 +29,78 @@
 
 namespace OfficeOpenXml.FormulaParsing.Excel.Functions.DateTime
 {
+	/// <summary>
+	/// Returns the appropriate day of the week (as an int) from the given date and return type if given valid input.
```
emdelaney: The "if given valid input" is implied and can be removed :)

EPPlus/FormulaParsing/Excel/Functions/DateTime/Weekday.cs
```diff
@@ -29,21 +29,78 @@
...
 	{
+		/// <summary>
+		/// Checks if input is valid, and returns the corresponding weekday value if so.
+		/// </summary>
+		/// <param name="arguments">The given arguments used to calculate the weekday.</param>
+		/// <param name="context">Unused in method, but necessary to override method.</param>
+		/// <returns>Returns the correct weekday number or an ExcelErrorValue, depending on if the input is valid.</returns>
 		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
 		{
 			if (ValidateArguments(arguments, 1) == false)
```
emdelaney: this.ValidateArguments

EPPlus/FormulaParsing/Excel/Functions/DateTime/Weekday.cs
```diff
@@ -29,21 +29,78 @@
...
 				return new CompileResult(eErrorType.Value);
+
+			var element1 = arguments.ElementAt(0).Value;
```
emdelaney: I'd rename this variable to reflect whatever the Excel name for this parameter is, like serialNumberCandidate.

EPPlus/FormulaParsing/Excel/Functions/DateTime/Weekday.cs
```diff
@@ -29,21 +29,78 @@
...
+					if((int)a == 46)
+						return new CompileResult(eErrorType.Value);
+				}
+			}
+				
+
+			if (arguments.Count() > 1 && arguments.ElementAt(1).Value is null)
+				return new CompileResult(eErrorType.Num);
+
+			if (arguments.Count() >1 && arguments.ElementAt(1).Value is string)
```
emdelaney: this can be an else if (). 

We should also add a space after the > sign.

EPPlus/FormulaParsing/Excel/Functions/DateTime/Weekday.cs
```diff
@@ -29,21 +29,78 @@
...
+
+			if(serialNumber < 0)
+				return new CompileResult(eErrorType.Num);
+
 			var returnType = arguments.Count() > 1 ? ArgToInt(arguments, 1) : 1;
-			return CreateResult(CalculateDayOfWeek(System.DateTime.FromOADate(serialNumber), returnType), DataType.Integer);
+
+			try
+			{
+				CalculateDayOfWeek(System.DateTime.FromOADate(serialNumber), returnType);
```
emdelaney: this.CalculateDayOfWeek

EPPlus/FormulaParsing/Excel/Functions/DateTime/Weekday.cs
```diff
@@ -29,21 +29,78 @@
...
+				return CreateResult(CalculateDayOfWeek(System.DateTime.FromOADate(serialNumber), returnType), DataType.Integer);
```
emdelaney: To make this easier to debug in the future, I'd break this into var result = this.CalculateDayOfWeek(); and then return this.CreateResult(result...).

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/WeekdayTests.cs
```diff
@@ -61,6 +62,62 @@ public void WeekdayShouldReturnCorrectResultForASundayWhenReturnTypeIs3()
 		}
 
 		[TestMethod]
+		public void WeekdayShouldReturnCorrectResultForASundayWhenReturnTypeIs11()
+		{
+			var func = new Weekday();
```
emdelaney: I would add an inline comment // to each of these test methods indicating what the case that's being tested is / what the return type means (// A return type of 11 requests the weekday as a 0-indexed value where the week starts on Tuesday, or whatever), so a future maintenance programmer has some hope of understanding why a test is failing.

EPPlus/FormulaParsing/Excel/Functions/DateTime/Weekday.cs
```diff
@@ -29,21 +29,78 @@
...
 				return new CompileResult(eErrorType.Value);
+
+			var element1 = arguments.ElementAt(0).Value;
+
+			if (element1 is null)
+				return new CompileResult(eErrorType.Num);
+
+			if (element1 is string)
+			{
+				foreach(var a in element1.ToString().ToCharArray())
```
emdelaney: You probably want to try var isDateString = DateTime.TryParse(element1, out DateTime date) here instead of checking the char array because there are other ways to write dates as strings (especially in other cultures) and rather than catch them all ourselves we'd prefer to let the C# System libraries do the work.

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/WeekdayTests.cs
```diff
@@ -69,6 +126,138 @@ public void WeekdayWithInvalidArgumentReturnsPoundValue()
...
+
+			var args = FunctionsHelper.CreateArgs("5/17/2017");
+			var result = func.Execute(args, this.ParsingContext);
+
+			Assert.AreEqual(4, result.Result);
+		}
+
+		[TestMethod]
+		public void WeekdayWithDateNotAsStringReturnsCorrectResult()
```
emdelaney: This test case can probably be removed since it's just testing args = CreateArgs(0).

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/WeekdayTests.cs
```diff
@@ -69,6 +126,138 @@ public void WeekdayWithInvalidArgumentReturnsPoundValue()
...
+		public void WeekdayWithDateAsStringWithDashesReturnsCorrectResult()
+		{
+			var func = new Weekday();
+
+			var args = FunctionsHelper.CreateArgs("5-17-2017");
+			var result = func.Execute(args, this.ParsingContext);
+
+			Assert.AreEqual(4, result.Result);
+		}
 		#endregion
```
emdelaney: I would also add a german-language test (I'll find your an example that uses a foreign-language-pattern) with the german date format 5.17.17.

EPPlus/FormulaParsing/Excel/Functions/DateTime/Weekday.cs
```diff
@@ -29,21 +29,78 @@
...
+			}
+				
+
+			if (arguments.Count() > 1 && arguments.ElementAt(1).Value is null)
+				return new CompileResult(eErrorType.Num);
+
+			if (arguments.Count() >1 && arguments.ElementAt(1).Value is string)
+				return new CompileResult(eErrorType.Value);
+
 			var serialNumber = ArgToDecimal(arguments, 0);
```
emdelaney: this.ArgToDecimal

EPPlus/FormulaParsing/Excel/Functions/DateTime/Weekday.cs
emdelaney: We should also add a space, after the foreach ().

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/WeekdayTests.cs
```diff
@@ -55,20 +62,252 @@ public void WeekdayShouldReturnCorrectResultForASundayWhenReturnTypeIs2()
...
+		}
+
+		[TestMethod]
+		public void WeekdayWithDateAsStringWithPeriodsReturnsCorrectResult()
+		{
+			// Test the case where the serial_number input is expressed as
+			// a date in a String.
+
+			var func = new Weekday();
+			var args = FunctionsHelper.CreateArgs("5.17.2017");
```
emdelaney: I'd be really tempted to test dd.mm.yy under a german culture with a day and a month that could be confused, like 5.6.17, and test that the weekday is still correct when run under German culture.

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/WeekdayTests.cs
```diff
@@ -33,12 +33,16 @@
...
+	[TestClass]
+	public class WeekdayTests : DateTimeFunctionsTestBase
 	{
 		#region Weekday Function (Execute) Tests
 		[TestMethod]
 		public void WeekdayShouldReturnCorrectResultForASundayWhenReturnTypeIs1()
 		{
+			// A return type of 1 requests the weekday as a 1-indexed value where
+			// the week starts on Sunday.
+
```
emdelaney: Hughes the vanquisher of Whitespace would appreciate if you vanquished this whitespace :)

EPPlus/FormulaParsing/Excel/Functions/DateTime/Weekday.cs
```diff
@@ -22,28 +22,90 @@
...
+					return new CompileResult(eErrorType.Num);
+				else if (returnTypeCandidate is string)
+				{
+					var isValidReturnType = Int32.TryParse(returnTypeCandidate.ToString(), out int result);
+					if (!isValidReturnType)
+						return new CompileResult(eErrorType.Value);
+				}
+
+			}
+
```
emdelaney: Remove all the empty lines above this line :)

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/WeekdayTests.cs
emdelaney: I think I'd actually leave this test as-is, because it is interesting undocumented behavior, and add a new test that tests it in german (because this should #VALUE under german dates since there is no month 17), and also in that new test test the valid (5.6.17, the 5th of June) case.

PR #11, opened 2017-05-18
Add comments to bring the Now Class and the NowTests Class up to the …

NaKister:
…coding standard.


EPPlus/FormulaParsing/Excel/Functions/DateTime/Now.cs
```diff
@@ -27,11 +27,20 @@
 
 namespace OfficeOpenXml.FormulaParsing.Excel.Functions.DateTime
 {
+	/// <summary>
+	/// A class that has the current date and time on the local computer.
```
emdelaney: Maybe "An Excel function that gets the current..."

PR #12, opened 2017-05-19
Bugs/699 improve support for day function

colewfrench:


EPPlus/FormulaParsing/Excel/Functions/DateTime/Day.cs
```diff
@@ -22,20 +22,47 @@
...
-			var dateObj = GetFirstValue(arguments);
-			var date = ParseDate(arguments, dateObj);
-
+			if (this.ValidateArguments(arguments, 1) == false)
+				return new CompileResult(eErrorType.Value);			
+			var serialNumberCandidate = this.GetFirstValue(arguments);
+			if (serialNumberCandidate is string)
+			{
+				var isValidDate = System.DateTime.TryParse(serialNumberCandidate.ToString(), out System.DateTime strDate);
+				var isValidNum = Double.TryParse(serialNumberCandidate.ToString(), out double dateAsNum);
```
emdelaney: I'd call this isValidNumber, and the double dateAsNumber.

EPPlus/FormulaParsing/Excel/Functions/DateTime/Day.cs
```diff
@@ -22,20 +22,47 @@
...
-				return new CompileResult(eErrorType.Value);
-			var dateObj = GetFirstValue(arguments);
-			var date = ParseDate(arguments, dateObj);
-
+			if (this.ValidateArguments(arguments, 1) == false)
+				return new CompileResult(eErrorType.Value);			
+			var serialNumberCandidate = this.GetFirstValue(arguments);
+			if (serialNumberCandidate is string)
+			{
+				var isValidDate = System.DateTime.TryParse(serialNumberCandidate.ToString(), out System.DateTime strDate);
```
emdelaney: maybe call this dateString instead of strDate.

EPPlus/FormulaParsing/Excel/Functions/DateTime/Day.cs
```diff
@@ -22,20 +22,47 @@
...
+				var isValidNum = Double.TryParse(serialNumberCandidate.ToString(), out double dateAsNum);
+				if (isValidNum)
+					serialNumberCandidate = dateAsNum;
+				else if (!isValidDate)
+					return new CompileResult(eErrorType.Value);
+			}
+			if (serialNumberCandidate is int)
```
emdelaney: There's some sweet C#7 syntax that lets us declare a variable as part of an is check, like `serialNumberCandidate is int serialNumber`, that saves us the casts after the check.

EPPlus/FormulaParsing/Excel/Functions/DateTime/Day.cs
```diff
@@ -22,20 +22,47 @@
...
+				serialNumberCandidate = ((int)serialNumberCandidate) * 1.0;
+			if (serialNumberCandidate is double)
```
emdelaney: You can use the sweet new C#7 syntax here, too.

EPPlus/FormulaParsing/Excel/Functions/DateTime/Day.cs
```diff
@@ -22,20 +22,47 @@
...
 		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
 		{
-			if (ValidateArguments(arguments, 1) == false)
-				return new CompileResult(eErrorType.Value);
-			var dateObj = GetFirstValue(arguments);
-			var date = ParseDate(arguments, dateObj);
-
+			if (this.ValidateArguments(arguments, 1) == false)
+				return new CompileResult(eErrorType.Value);			
+			var serialNumberCandidate = this.GetFirstValue(arguments);
```
emdelaney: This sequence is basically a better ParseDate method, right? I think we should refactor this date-parsing code out into its own method so we can re-use it (and replace / eliminate ParseDate at the same time).

EPPlus/FormulaParsing/Excel/Functions/DateTime/Day.cs
```diff
@@ -22,20 +22,47 @@
...
+			if (serialNumberCandidate is int)
+				serialNumberCandidate = ((int)serialNumberCandidate) * 1.0;
+			if (serialNumberCandidate is double)
+			{
+				if ((double)serialNumberCandidate == 0)
+					return CreateResult(0, DataType.Integer);
+				else if ((double)serialNumberCandidate < 0)
+					return new CompileResult(eErrorType.Num);
+			}
+			var date = this.ParseDate(arguments, serialNumberCandidate);
```
emdelaney: Expanding on my previous comment, I'd eliminate the base class that has the ParseDate call, move it to ConvertUtil (possibly as `internal static bool TryParseDateObject(object dateCandidate, out DateTime date)`)and then call that instead. I'd also add tests that directly test that method so we can be confident in it.

emdelaney: Then this Execute method would become validate input -> call conversion utility -> return error code or date.Day, which would be very clean.

EPPlus/FormulaParsing/Excel/Functions/DateTime/Day.cs
```diff
@@ -22,20 +22,47 @@
...
 			return CreateResult(date.Day, DataType.Integer);
```
emdelaney: this.CreateResult

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/DayTests.cs
```diff
@@ -43,27 +43,31 @@ public class DayTests : DateTimeFunctionsTestBase
 	{
 		#region Day Function (Execute) Tests
```
emdelaney: All of these added tests are great <3

EPPlus/FormulaParsing/Excel/Functions/DateTime/Day.cs
```diff
@@ -44,27 +45,14 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
...
-				if (isValidNum)
-					serialNumberCandidate = dateAsNumber;
-				else if (!isValidDate)
-					return new CompileResult(eErrorType.Value);
-			}
+			var isValidDate = ConvertUtil.TryParseDateObject(serialNumberCandidate, out System.DateTime date, out eErrorType? error);
+			if (isValidDate)
+				return this.CreateResult(date.Day, DataType.Integer);
 			if (serialNumberCandidate is int serialNumberInt)
 				serialNumberCandidate = serialNumberInt * 1.0;
```
emdelaney: this can be simplified to (double)serialNumberInt;

EPPlus/Utils/ConvertUtil.cs
```diff
@@ -65,6 +65,45 @@ internal static bool TryParseDateString(object candidate, out DateTime result)
...
+			{
+				var isValidDate = DateTime.TryParse(dateCandidate.ToString(), out date);
+				if (isValidDate)
+					return true;
+				var isValidNumber = Double.TryParse(dateCandidate.ToString(), out double dateAsNumber);
+				if (isValidNumber)
+					dateCandidate = dateAsNumber;
+			}
+			if (dateCandidate is int dateInt)
+				dateCandidate = dateInt * 1.0;
```
emdelaney: We can cast this as a double(dateInt);

EPPlusTest/Utils/ConvertUtilTest.cs
```diff
@@ -115,5 +116,36 @@ public void GetValueDouble()
 			Assert.AreEqual(0d, ConvertUtil.GetValueDouble("Not a number"));
 			Assert.AreEqual(double.NaN, ConvertUtil.GetValueDouble("Not a number", false, true));
 		}
+
+		[TestMethod]
+		public void TestTryParseObjectString()
```
emdelaney: I'd make this its own region (TryParseDateObject Tests) and then make each test case its own test (TryParseDateObjectParsesInt, ...ParsesDouble, etc).

EPPlus/FormulaParsing/Excel/Functions/DateTime/Day.cs
```diff
@@ -22,21 +22,37 @@
...
-			var date = ParseDate(arguments, dateObj);
-
-			return CreateResult(date.Day, DataType.Integer);
+			if (this.ValidateArguments(arguments, 1) == false)
+				return new CompileResult(eErrorType.Value);			
+			var serialNumberCandidate = this.GetFirstValue(arguments);
+			var isValidDate = ConvertUtil.TryParseDateObject(serialNumberCandidate, out System.DateTime date, out eErrorType? error);
+			if (isValidDate)
+				return this.CreateResult(date.Day, DataType.Integer);
+			if (serialNumberCandidate is int serialNumberInt)
```
emdelaney: I'd be tempted to move these two cases above the date check and chain them together, if((serialNumberCAndidate is int... && serialNumberInt == 0) || <linebreak> (serialNumberCandidate is double.. && snDouble == 0.0)) { return 0;} with a comment indicating that zero is special-cased.

EPPlus/FormulaParsing/Excel/Functions/DateTime/Day.cs
```diff
@@ -22,21 +22,37 @@
...
+	/// Returns the corresponding day of the month (as an int) from the given date.
+	/// </summary>
 	public class Day : DateParsingFunction
 	{
+		/// <summary>
+		/// Checks if the input is valid, and returns the corresponding day of the month if so.
+		/// </summary>
+		/// <param name="arguments">The given arguments used to calculate the day of the month.</param>
+		/// <param name="context">Unused in the method, but necessary to override the method.</param>
+		/// <returns></returns>
```
emdelaney: Don't forget to fill in the return value!

EPPlus/FormulaParsing/Excel/Functions/DateTime/Day.cs
```diff
@@ -22,21 +22,37 @@
  *******************************************************************************
  * Mats Alm   		                Added		                2013-12-03
  *******************************************************************************/
+using System;
+using OfficeOpenXml.Utils;
```
emdelaney: These should be sorted (right click -> remove and sort usings).

PR #18, opened 2017-05-23
Bugs/699 improve support for year function

colewfrench:


EPPlus/FormulaParsing/Excel/Functions/DateTime/Year.cs
```diff
@@ -25,27 +25,34 @@
 using System.Collections.Generic;
 using System.Linq;
 using OfficeOpenXml.FormulaParsing.ExpressionGraph;
+using OfficeOpenXml.Utils;
 
 namespace OfficeOpenXml.FormulaParsing.Excel.Functions.DateTime
 {
+	/// <summary>
+	/// Returns the corresponding year of the given date as an int.
```
emdelaney: "as an int" can be removed.

EPPlus/FormulaParsing/Excel/Functions/DateTime/Year.cs
```diff
@@ -25,27 +25,34 @@
...
+				return this.CreateResult(1900, DataType.Integer);
+			if (ConvertUtil.TryParseDateObject(dateObj, out System.DateTime date, out eErrorType? error))
+				return this.CreateResult(date.Year, DataType.Integer);
 			else
-			{
-				var d = ArgToDecimal(arguments, 0);
-				date = System.DateTime.FromOADate(d);
-			}
-			return CreateResult(date.Year, DataType.Integer);
+				return new CompileResult((eErrorType)error);
```
emdelaney: Consider error.Value instead.

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/YearTests.cs
```diff
@@ -100,6 +99,165 @@ public void YearFunctionWorksInDifferentCultureDateFormats()
...
+			Assert.AreEqual(2017, result.Result);
+		}
+
+		[TestMethod]
+		public void YearWithDateAsDoubleReturnsCorrectYear()
+		{
+			// Note that 42874.34114 is the OADate representation of
+			// some time on May 19, 2017.
+			var func = new Year();
+			var args = FunctionsHelper.CreateArgs(42874.34114);
```
emdelaney: I like the pattern from the deleted test before where they create the DateTime for the day and then create the args by using the ToOADate() call, it makes it more obvious which day is being tested. I'd do that here by creating a DateTime and adding some time to it, and then calling ToOADate().

PR #30, opened 2017-05-31
Bugs/706 improve tests for time function

NaKister:
Added a new function called TryArgToInt that attempts to parse an argument to an integer value. I also cleaned up the Time class with the use of this new function.


EPPlus/FormulaParsing/Excel/Functions/DateTime/Time.cs
```diff
@@ -42,33 +42,38 @@ public class Time : TimeBaseFunction
 		/// <returns>The time as a double (decimal numer).</returns>
 		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
 		{
-			double dateResult;
+			int value;
+			eErrorType? err;
```
emdelaney: this should be a full word like error / errorCode / errorType

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -145,19 +144,71 @@ private bool TooFewArgs(IEnumerable<FunctionArgument> arguments, int minLength)
...
+		/// </summary>
+		/// <param name="arguments">The list of function arguments where our input to parse is.</param>
+		/// <param name="index"> The index of the arguments to try to parse to an integer.</param>
+		/// <param name="value">The resulting value if the parse was successful. If not the value is -100.</param>
+		/// <param name="err">Null if parse was successful, or the <see cref="eErrorType"/> indicating why the parse was unsuccessful.</param>
+		/// <returns></returns>
+		protected bool TryArgToInt(IEnumerable<FunctionArgument> arguments, int index, out int value, out eErrorType? err)
+		{
+		//	throw new NotImplementedException();
+			var val = arguments.ElementAt(index).Value;
```
emdelaney: value

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -145,19 +144,71 @@ private bool TooFewArgs(IEnumerable<FunctionArgument> arguments, int minLength)
...
+		/// Attempts to parse an argument into an integer value.
+		/// </summary>
+		/// <param name="arguments">The list of function arguments where our input to parse is.</param>
+		/// <param name="index"> The index of the arguments to try to parse to an integer.</param>
+		/// <param name="value">The resulting value if the parse was successful. If not the value is -100.</param>
+		/// <param name="err">Null if parse was successful, or the <see cref="eErrorType"/> indicating why the parse was unsuccessful.</param>
+		/// <returns></returns>
+		protected bool TryArgToInt(IEnumerable<FunctionArgument> arguments, int index, out int value, out eErrorType? err)
+		{
+		//	throw new NotImplementedException();
```
emdelaney: Delete this :)

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -145,19 +144,71 @@ private bool TooFewArgs(IEnumerable<FunctionArgument> arguments, int minLength)
...
 
 		/// <summary>
+		/// Attempts to parse an argument into an integer value.
+		/// </summary>
+		/// <param name="arguments">The list of function arguments where our input to parse is.</param>
+		/// <param name="index"> The index of the arguments to try to parse to an integer.</param>
+		/// <param name="value">The resulting value if the parse was successful. If not the value is -100.</param>
+		/// <param name="err">Null if parse was successful, or the <see cref="eErrorType"/> indicating why the parse was unsuccessful.</param>
+		/// <returns></returns>
+		protected bool TryArgToInt(IEnumerable<FunctionArgument> arguments, int index, out int value, out eErrorType? err)
```
emdelaney: The C# Pattern for this is technically called the "TryGet" pattern, so we should follow that by right click -> refactor -> rename-ing this to something that starts with TryGet, like TryGetArgAsInt or TryGetIntFromArg (I like the former better, personally, but I'm not attached to either).

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -145,19 +144,71 @@ private bool TooFewArgs(IEnumerable<FunctionArgument> arguments, int minLength)
...
+		{
+		//	throw new NotImplementedException();
+			var val = arguments.ElementAt(index).Value;
+			err = null;
+			value = -100;
```
emdelaney: I'd set this to zero or int.MinValue if it's just a default "unparsed" value.

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -145,19 +144,71 @@ private bool TooFewArgs(IEnumerable<FunctionArgument> arguments, int minLength)
...
+
+			if (val == null)
+			{
+				value = 0;
+				return true;
+			}
+			if (val is int)
```
emdelaney: else if?

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -145,19 +144,71 @@ private bool TooFewArgs(IEnumerable<FunctionArgument> arguments, int minLength)
...
+			{
+				value = this.ArgToInt(arguments, index);
+				return true;
+			}
+			if (val is double)
```
emdelaney: else if?

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -145,19 +144,71 @@ private bool TooFewArgs(IEnumerable<FunctionArgument> arguments, int minLength)
...
+			{
+				value = this.ArgToInt(arguments, index);
+				return true;
+			}
+			if (val is string)
```
emdelaney: else if?

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -145,19 +144,71 @@ private bool TooFewArgs(IEnumerable<FunctionArgument> arguments, int minLength)
...
+					value = this.ArgToInt(arguments, index);
+					return true;
+				}
+				else if (Utils.ConvertUtil.TryParseDateString(val, out System.DateTime resu))
+				{
+					err = eErrorType.Num;
+					return false;
+				}
+			}
+			if (Utils.ConvertUtil.TryParseDateObject(val, out System.DateTime re, out eErrorType? error))
```
emdelaney: NOT an else if :)

emdelaney: Also this can just be ConvertUtil.TryParse... with a using (control-. to get the handy error-fixing context menu).

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -145,19 +144,71 @@ private bool TooFewArgs(IEnumerable<FunctionArgument> arguments, int minLength)
...
+				return true;
+			}
+			if (val is string)
+			{
+				if (Utils.ConvertUtil.TryParseNumericString(val, out double result))
+				{
+					value = this.ArgToInt(arguments, index);
+					return true;
+				}
+				else if (Utils.ConvertUtil.TryParseDateString(val, out System.DateTime resu))
```
emdelaney: I believe this else-if needs to go away, because these functions treat small dates just fine ("Jan 30, 1900" gets treated like 30, for example-- the #NUM! comes from the out-of-bounds check on the huge date 42000+ that corresponds to 2017.

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
emdelaney: Also re should be renamed to something like date.

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -145,19 +144,71 @@ private bool TooFewArgs(IEnumerable<FunctionArgument> arguments, int minLength)
...
+				{
+					err = eErrorType.Num;
+					return false;
+				}
+			}
+			if (Utils.ConvertUtil.TryParseDateObject(val, out System.DateTime re, out eErrorType? error))
+			{
+				value = this.ArgToInt(arguments, index);
```
emdelaney: this should probably be (int)date.ToOADate();

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/TimeTests.cs
```diff
@@ -100,19 +100,19 @@ public void TimeWithFirstArgNegativeReturnsPoundNum()
 			var function = new Time();
 
 			var case1Args = FunctionsHelper.CreateArgs(-10, 10, 10);
-			var case2Args = FunctionsHelper.CreateArgs(-10, 10, -10);
-			var case3Args = FunctionsHelper.CreateArgs(-10, -10, 10);
-			var case4Args = FunctionsHelper.CreateArgs(-10, -10, -10);
+			//var case2Args = FunctionsHelper.CreateArgs(-10, 10, -10);
```
emdelaney: These should be uncommented or removed.

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/TimeTests.cs
```diff
@@ -100,19 +100,19 @@ public void TimeWithFirstArgNegativeReturnsPoundNum()
...
+			//var case3Args = FunctionsHelper.CreateArgs(-10, -10, 10);
+			//var case4Args = FunctionsHelper.CreateArgs(-10, -10, -10);
```
emdelaney: We should also add a new unit test with a small date string (like "1/30/1900") that shows that it gets treated as its OADate of 30.

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
NaKister: If I take that else-if out inputs as dates don't return #NUM! like they are supposed to. I did some checking and any date as a string that is put into the TIME function needs to return a #NUM!

NaKister: Nevermind- I moved some stuff around and got it to work without the else-if there.

emdelaney: Hmm. The test case ```=TIME("1/4/1900","1/24/1900","1/13/1900")``` seems to treat the string dates as OADates?

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -159,47 +160,42 @@ protected int ArgToInt(IEnumerable<FunctionArgument> arguments, int index)
...
-				{
-					err = eErrorType.Num;
-					return false;
-				}
 			}
-			if (Utils.ConvertUtil.TryParseDateObject(val, out System.DateTime re, out eErrorType? error))
+			if (ConvertUtil.TryParseDateObject(intCandidate, out System.DateTime date, out eErrorType? error))
 			{
-				value = this.ArgToInt(arguments, index);
+				var testVal = date.ToOADate();
```
emdelaney: Does this need to exist?

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/TimeTests.cs
```diff
@@ -320,6 +320,16 @@ public void TimeWithDateReturnsPoundNum()
 			var result = function.Execute(args, this.ParsingContext);
 			Assert.AreEqual(eErrorType.Num, ((ExcelErrorValue)result.Result).Type);
 		}
+
+		[TestMethod]
+		public void TimeWithSmallDateStringReturnsPoundNum()
```
emdelaney: This test doesn't actually return #NUM!, so we should rename it.

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -115,7 +116,7 @@ protected object GetFirstValue(IEnumerable<FunctionArgument> val)
 		/// <param name="minLength"></param>
 		protected bool ValidateArguments(IEnumerable<FunctionArgument> arguments, int minLength)
 		{
-			Require.That(arguments).Named("arguments").IsNotNull();
+			Utilities.Require.That(arguments).Named("arguments").IsNotNull();
```
emdelaney: We should fix this in a separate branch / pull request after this one, but this line should eventually be `if (arguments == null) return false;` instead of this functional approach. This may break a bunch of stuff, though, if there are poorly-implemented functions.

PR #31, opened 2017-05-31
Bugs/65 data validation row expansion

Tauroctonos:
Added an update function and placed it at the end of the InsertRow and InsertColumn functions. Data Validation List ranges should now properly expand when you add rows and columns inside of them.

### DIFF
--- a/EPPlus/ExcelWorksheet.cs
+++ b/EPPlus/ExcelWorksheet.cs
@@ -4761,13 +4761,13 @@ namespace OfficeOpenXml
 					var validation = sheet.DataValidations.ElementAt(i) as DataValidation.Contracts.IExcelDataValidationList;
 					if(validation != null)
 					{
-						
-						if (validation.Address.WorkSheet == null ||
-						validation.Address.WorkSheet == this.Name.ToUpper())
-						{
-							var newFormula = this.Package.FormulaManager.UpdateFormulaReferences(validation.Formula.ExcelFormula, rows, columns, rowFrom, columnFrom, sheet.Name, sheet.Name);
+						string newFormula = "!";
+						if (validation.Address.WorkSheet == null ) //This formula references the sheet it is on
+							newFormula = this.Package.FormulaManager.UpdateFormulaReferences(validation.Formula.ExcelFormula, rows, columns, rowFrom, columnFrom, sheet.Name, sheet.Name);
+						else if(validation.Address.WorkSheet == this.Name.ToUpper()) //This formula references another sheet in the workbook
+							newFormula = this.Package.FormulaManager.UpdateFormulaReferences(validation.Formula.ExcelFormula, rows, columns, rowFrom, columnFrom, validation.Address.WorkSheet, validation.Address.WorkSheet);
+						if(!newFormula.Equals("!")) //Only update the formula if we've succesfully created an updated formula
 							validation.Formula.ExcelFormula = newFormula;
-						}
 					}
 				}
 			}
--- a/EPPlus/ExcelWorksheet.cs
+++ b/EPPlus/ExcelWorksheet.cs
@@ -1425,6 +1425,7 @@ namespace OfficeOpenXml
 			{
 				sheet.UpdateCrossSheetReferences(this.Name, rowFrom, rows, 0, 0);
 			}
+			this.UpdateDataValidationRanges(rowFrom, rows, 0, 0);
 		}
 
 		/// <summary>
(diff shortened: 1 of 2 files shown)


EPPlus/ExcelWorksheet.cs
```diff
@@ -4749,6 +4751,28 @@ private XmlNode CreateHyperLinkCollection()
...
+			{
+				for (int i = sheet.DataValidations.Count - 1; i >= 0; i--)
+				{
+					//Update ListValidations
+					var validation = sheet.DataValidations.ElementAt(i) as DataValidation.Contracts.IExcelDataValidationList;
+					if(validation != null)
+					{
+						
+						if (validation.Address.WorkSheet == null ||
+						validation.Address.WorkSheet == this.Name.ToUpper())
```
emdelaney: Does the WorkSheet also need to be ToUpper()-ed?

EPPlusTest/ExcelWorksheetTests.cs
```diff
@@ -3176,6 +3176,67 @@ public void InsertRowAndColumnsInsideChartUpdatesChartTo()
...
+				var validation = sheetValidations.DataValidations.AddListValidation(@"'Sheet'!" + sheetTarget.Cells["D5"].Address);
+				validation.Formula.ExcelFormula = "=Sheet!$B$2:$E$5";
+
+				//expand the range
+				sheetTarget.InsertColumn(3, 2);
+				sheetTarget.InsertRow(3, 2);
+
+				//validate that the Data Validation range has also expanded
+				var validationRange = sheetValidations.DataValidations.First() as OfficeOpenXml.DataValidation.Contracts.IExcelDataValidationList;
+				Assert.AreEqual("=SHEET!$B$2:$E$5", validationRange.Formula.ExcelFormula);
```
emdelaney: This expected value look un-expanded? Either the comment or the expected value should be updated here.

EPPlus/ExcelWorksheet.cs
```diff
@@ -4749,6 +4751,28 @@ private XmlNode CreateHyperLinkCollection()
 			return _worksheetXml.DocumentElement.InsertAfter(hl, prevNode);
 		}
 
+		private void UpdateDataValidationRanges(int rowFrom, int rows, int columnFrom, int columns)
+		{
+			foreach(var sheet in this.Workbook.Worksheets)
+			{
+				for (int i = sheet.DataValidations.Count - 1; i >= 0; i--)
+				{
+					//Update ListValidations
```
emdelaney: Our coding standard suggests that this comment should go away :\

EPPlus/ExcelWorksheet.cs
```diff
@@ -4749,6 +4751,28 @@ private XmlNode CreateHyperLinkCollection()
...
+			{
+				for (int i = sheet.DataValidations.Count - 1; i >= 0; i--)
+				{
+					//Update ListValidations
+					var validation = sheet.DataValidations.ElementAt(i) as DataValidation.Contracts.IExcelDataValidationList;
+					if(validation != null)
+					{
+						string newFormula = "!";
+						if (validation.Address.WorkSheet == null ) //This formula references the sheet it is on
+							newFormula = this.Package.FormulaManager.UpdateFormulaReferences(validation.Formula.ExcelFormula, rows, columns, rowFrom, columnFrom, sheet.Name, sheet.Name);
```
emdelaney: This looks like it will incorrectly update the value of a validation range in the test case (which we should add a test for) where there is a validation reference without a specified worksheet (implicit same-sheet) on sheet A, and we add rows to sheet B. We'd expect no change to the local references on sheet A.

emdelaney: UpdateFormulaReferences also takes a current sheet and a modified sheet-- we should be passing `this` in to the modified sheet, which should cover this new test case. We might even be able to delete the else if case below by passing in the correct sheets to UpdateFormulaReferences.

EPPlusTest/ExcelWorksheetTests.cs
```diff
@@ -3176,6 +3176,67 @@ public void InsertRowAndColumnsInsideChartUpdatesChartTo()
...
+				//expand the range
+				sheetTarget.InsertColumn(3, 2);
+				sheetTarget.InsertRow(3, 2);
+
+				//validate that the Data Validation range has also expanded
+				var validationRange = sheetValidations.DataValidations.First() as OfficeOpenXml.DataValidation.Contracts.IExcelDataValidationList;
+				Assert.AreEqual("='SHEET'!$B$2:$G$7", validationRange.Formula.ExcelFormula);
+			}
+		}
+
```
emdelaney: We should add more test cases to get comprehensive coverage of the logic:

- A validation range on sheet A that references sheet B:
    - Insert rows on Sheet A -> No change
    - Insert rows on Sheet B -> Reference is updated
- An implicit validation range: 
    - Copy the first two existing first two tests, but without SHEET! specified. (Verify that you can create this at all in Excel, it may be an invalid test case).
    - Test that an implicit validation range on sheet A is not updated when rows are added to sheet B
- Figure out if there are any nuances around relative or absolute references and add tests as appropriate:
    - If it's possible to put in relative references, then we should duplicate these tests against relative references.

EPPlus/ExcelWorksheet.cs
Tauroctonos: Noted and deleted

EPPlus/ExcelWorksheet.cs
Tauroctonos: CurrentSheet is supposed to be the sheet that is referenced by the formula we're updating. We need the if/else because if the sheet we're adding rows and columns to (this) is not the sheet with the Data Validation, then the 'current sheet' is stored in that validation's Address.Worksheet .

You're definitely right that the modified sheet should be from 'this'.

EPPlusTest/ExcelWorksheetTests.cs
Tauroctonos: I've added tests for implicit ranges, as well as some checks to make sure we only update ranges that we're actually adding columns/rows to. Since implicit/relative references only reference the same sheet, there were only the two tests that needed to be duplicated.

PR #33, opened 2017-05-31
Fixes doughnut chart null reference exception (vsts bug 632).

EvanSchallerer:


EPPlus/Drawing/Chart/ExcelDoughnutChart.cs
```diff
@@ -110,7 +110,7 @@ internal override eChartType GetChartType(string name)
 		{
 			if (name == "doughnutChart")
 			{
-				if (((ExcelPieChartSerie)Series[0]).Explosion > 0)
+				if (this.Series?.Count > 0 && ((ExcelPieChartSerie)Series[0]).Explosion > 0)
```
emdelaney: This was an error reading a workbook that had an exploded pie chart, right? We should add a test case like the other chart test cases where we check in a workbook with an exploded pie chart and try to read it.

PR #36, opened 2017-06-01
Bugs/700 add tests for days306 function

NaKister:
Added support for a missing test case.


EPPlus/FormulaParsing/Excel/Functions/DateTime/Days360.cs
```diff
@@ -34,6 +34,12 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
 			if (arguments.Count() > 2)
 			{
 				var european = false;
+				if (arguments.ElementAt(2).Value is string stringVal)
+				{
+					if (!stringVal.Equals("true") && !stringVal.Equals("false"))
```
emdelaney: You're going to want to feed in some extra parameters to .Equals here to test upper- and lower-case values, probably currentcultureignorecase.

PR #44, opened 2017-06-02
Bugs/702 testing eomonth function

NaKister:
Added support for two missing test cases.

### DIFF
--- a/EPPlus/FormulaParsing/Excel/Functions/DateTime/Eomonth.cs
+++ b/EPPlus/FormulaParsing/Excel/Functions/DateTime/Eomonth.cs
@@ -90,7 +90,7 @@ namespace OfficeOpenXml.FormulaParsing.Excel.Functions.DateTime
 				resultDate = new System.DateTime(date.Year, date.Month, 1).AddMonths(monthsToAdd + 1).AddDays(-2);
 			}
 
-			return CreateResult(resultDate.ToOADate(), DataType.Date);
+			return this.CreateResult(resultDate.ToOADate(), DataType.Date);
 		}
 
 		#endregion
--- a/EPPlus/FormulaParsing/Excel/Functions/DateTime/Eomonth.cs
+++ b/EPPlus/FormulaParsing/Excel/Functions/DateTime/Eomonth.cs
@@ -38,8 +38,6 @@ namespace OfficeOpenXml.FormulaParsing.Excel.Functions.DateTime
 				if (intResult == 0 || intResult == 0.0)
 					date = System.DateTime.FromOADate(result + 2);
 			}
-			else if (dateArgument.ToString().Equals("2/29/1900") || dateArgument.ToString().Equals("2.29.1900") || dateArgument.ToString().Equals("2-29-1900"))
-				date = System.DateTime.FromOADate(2);
 			else
 				return new CompileResult(eErrorType.Value);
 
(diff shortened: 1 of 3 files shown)


EPPlus/FormulaParsing/Excel/Functions/DateTime/Eomonth.cs
```diff
@@ -29,14 +29,17 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
...
 				if (result < 0)
 					return new CompileResult(eErrorType.Num);
-				date = System.DateTime.FromOADate(result);
-				if (result == 0 || result == 0.0)
+				int intResult = (int)result;
+				date = System.DateTime.FromOADate(intResult);
+				if (intResult == 0 || intResult == 0.0)
 					date = System.DateTime.FromOADate(result + 2);
 			}
+			else if (dateArgument.ToString().Equals("2/29/1900") || dateArgument.ToString().Equals("2.29.1900") || dateArgument.ToString().Equals("2-29-1900"))
```
emdelaney: this... Oh, yeah, we can't parse this, can we. Ooof. Good catch, I suppose. We should think about if we can do this more generally, or if this should be a known error case (I'd be tempted to say it's a known-not-supported case), because this solution won't work for other cultures.

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/EomonthTests.cs
```diff
@@ -175,6 +175,27 @@ public void EomonthOnFebruary291900ReturnsFebruary28()
 			Assert.AreEqual(59d, result.Result);
 		}
 
+		[TestMethod]
+		public void EomonthOnFeb291900AsDateReturnsFebruary28()
```
emdelaney: After talking with Hughes, we agree that we shouldn't actuallysupport this case. We should, however, add unit tests that demonstrate what happens in EPPlus when you try it, with a prominent comment in the test that it's known-different behavior from Excel.

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/EomonthTests.cs
```diff
@@ -175,6 +175,27 @@ public void EomonthOnFebruary291900ReturnsFebruary28()
...
+			Assert.AreEqual(59d, result.Result);
+		}
+
+		[TestMethod]
+		public void EomonthDateNotAsStringInExcelWorkbookReturnsCorrectValue()
+		{
+			using (var package = new ExcelPackage())
+			{
+				var ws = package.Workbook.Worksheets.Add("Sheet1");
+				ws.Cells["B2"].Formula = "EOMONTH(5/15/2017, 1)";
```
emdelaney: This test case is probably unnecessary since it's basically EOMONTH(0.3ish, 1).

EPPlusTest/FormulaParsing/Excel/Functions/DateTimeFunctions/EomonthTests.cs
```diff
@@ -180,20 +180,15 @@ public void EomonthOnFeb291900AsDateReturnsFebruary28()
 		{
 			var function = new Eomonth();
 			var result = function.Execute(FunctionsHelper.CreateArgs("2/29/1900", 1), this.ParsingContext);
```
emdelaney: We should add a comment here indicating that this behavior is different from Excel because the day 2/29/1900 never actually happened.

PR #45, opened 2017-06-05
Bugs/791 ensure all date time functions have header

NaKister:
Added headers and regions to classes that were missing them.


EPPlus/FormulaParsing/Excel/Functions/DateTime/Weekday.cs
```diff
@@ -1,27 +1,29 @@
...
- * If you unfamiliar with this license or have questions about it, here is an http://www.gnu.org/licenses/gpl-faq.html
- *
- * All code and executables are provided "as is" with no warranty either express or implied. 
- * The author accepts no liability for any damage or loss of business that this product may cause.
- *
- * Code change notes:
- * 
- * Author							Change						Date
- *******************************************************************************
- * Mats Alm   		                Added		                2013-12-03
```
emdelaney: For any of these that have a Mats comment, we should leave the Mats author line and add another line after saying to look at the source control history.

PR #46, opened 2017-06-06
Bugs/805 support product function

NaKister:
Added support for the product function as well as added unit tests for the multiplication operator.


EPPlusTest/FormulaParsing/Excel/Functions/Math/ProductTests.cs
```diff
@@ -0,0 +1,488 @@
...
+		}
+
+		[TestMethod]
+		public void ProductWithMaxInputsReturnsCorrectValue()
+		{
+			// The maximum number of inputs the function takes is 264.
+			using (var package = new ExcelPackage())
+			{
+				var ws = package.Workbook.Worksheets.Add("Sheet1");
+				ws.Cells["B1"].Value = 2;
```
emdelaney: You might be able to shrink this down by initializing these in a loop and using the Cells[row, column] access notation.

EPPlusTest/FormulaParsing/Excel/OperatorsTests.cs
```diff
@@ -245,11 +271,371 @@ public void OperatorMultiplyShouldThrowExceptionIfNonNumericOperand()
...
+		}
+
+		[TestMethod]
+		public void OperatorMultiplyWithMaxInputsReturnsCorrectValue()
+		{
+			// The maximum number of inputs the function takes is 264.
+			using (var package = new ExcelPackage())
+			{
+				var ws = package.Workbook.Worksheets.Add("Sheet1");
+				ws.Cells["B1"].Value = 2;
```
emdelaney: Same here, you might be able to shrink this down with a for-loop.

EPPlus/FormulaParsing/Excel/Functions/Math/Product.cs
```diff
@@ -1,42 +1,71 @@
...
 		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
 		{
 			if (ValidateArguments(arguments, 1) == false)
 				return new CompileResult(eErrorType.Value);
 			var result = 0d;
 			var index = 0;
+			double valAsDouble;
+			foreach (var val in arguments)
+			{
+				if (val.Value is int || val.Value is double) { }
```
emdelaney: Is this exactly the same logic that the Multiply operator does? If so, we should centralize it somewhere and expose it so both classes are calling the same method, instead of having identical copies of the code.

PR #47, opened 2017-06-06
Bugs/747 ensure date functions consider zero date

colewfrench:


EPPlusTest/Utils/ConvertUtilTest.cs
```diff
@@ -213,11 +222,77 @@ public void TryParseDateObjectToOADateParsesDateAsString()
...
+				}
+			}
+			finally
+			{
+				Thread.CurrentThread.CurrentCulture = currentCulture;
+			}
+		}
+
+		[TestMethod]
+		public void TryParseDateObjectAsOADateDoesNotParseGermanStyleDatesAsDoubles()
```
emdelaney: This test needs a new, correct name and a big, obvious comment indicating that this is NOT what excel does.

EPPlusTest/Utils/ConvertUtilTest.cs
```diff
@@ -213,11 +222,77 @@ public void TryParseDateObjectToOADateParsesDateAsString()
...
-		public void TryParseDateObjectToOADateDoesNotParseNonDateString()
-		{
-			var isValidDate = ConvertUtil.TryParseDateObjectToOADate("word", out double OADate);
-			Assert.AreEqual(false, isValidDate);
-			Assert.AreEqual(-1.0, OADate);
+		public void TryParseDateObjectToOADateParsesDoublesInStringsAsDoublesCorrectly()
+		{
+			var testNumber = "1.11";
+			var isValidOADate = ConvertUtil.TryParseDateObjectToOADate(testNumber, out double OADate);
+			Assert.AreEqual(1.11, OADate);
```
emdelaney: We should also assert that isValidOADate is true.

PR #49, opened 2017-06-07
Bugs/747 ensure date functions consider zero date

colewfrench:


EPPlus/FormulaParsing/Excel/Functions/DateTime/Days360.cs
```diff
@@ -53,10 +53,27 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
 		{
 			if (this.ValidateArguments(arguments, 2) == false)
 				return new CompileResult(eErrorType.Value);
-			if (!this.TryGetArgumentDateValueAtIndex(arguments, 0, out System.DateTime dt1) || !this.TryGetArgumentDateValueAtIndex(arguments, 1, out System.DateTime dt2))
+			var dateObj1 = arguments.ElementAt(0).Value;
```
emdelaney: If you can think of non-numeric ways to describe these (startDateObject, startOADate, etc) that would be preferable.

PR #50, opened 2017-06-07
Fixed an issue with ValueShouldHandleTime()

NAccuardi:
Passes all tests again.


EPPlus/FormulaParsing/Excel/Functions/Text/Value.cs
```diff
@@ -42,7 +42,7 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
 			if (Regex.IsMatch(val, @"^[\d]{1,2}" + timeSeparator + @"[\d]{2}(" + timeSeparator + @"[\d]{2})?$"))
 			{
 				var timeResult = _timeValueFunc.Execute(val);
-				if (timeResult.DataType == DataType.Date)
+				if (timeResult.DataType == DataType.Decimal)
```
emdelaney: Would it be better to do, like, not-equal-to-excelError?

PR #53, opened 2017-06-08
Bugs/799 improve support for degrees and radians functions

colewfrench:


EPPlusTest/FormulaParsing/Excel/Functions/Math/DegreesTests.cs
```diff
@@ -0,0 +1,175 @@
...
+*******************************************************************************/
+using EPPlusTest.FormulaParsing.TestHelpers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using OfficeOpenXml;
+using OfficeOpenXml.FormulaParsing.Excel.Functions.Math;
+
+namespace EPPlusTest.FormulaParsing.Excel.Functions.Math
+{
+	[TestClass]
+	public class DegreesTests : MathFunctionsTestBase
```
emdelaney: I'd add one test to each of these that uses a value outside of the unit circle, like 720 degrees or 4pi radians.

EPPlus/FormulaParsing/Excel/Functions/Math/Degrees.cs
```diff
@@ -1,17 +1,58 @@
...
+		/// <returns>Returns the given radian value converted to degrees.</returns>
 		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
 		{
-			if (ValidateArguments(arguments, 1) == false)
+			if (this.ValidateArguments(arguments, 1) == false)
+				return new CompileResult(eErrorType.Value);
+			var angleObj = arguments.ElementAt(0).Value;
+			// The method TryParseDateObjectToOADate attempts to parse the given object as a double.
+			// That double is only used as an OADate if the given object represented a date.
+			if (!ConvertUtil.TryParseDateObjectToOADate(angleObj, out double angle))
```
emdelaney: Question: should we think about renaming this convertutil method eventually (not right now, necessarily)? It looks like it's basically an ObjectToDecimal parser at this point, a la ArgToDecimal.

PR #55, opened 2017-06-09
Add checks for PivotTable cache definition types before attempting to update ranges

ZacharyPF:
Fix issue where NullReferenceExceptions occurred because PivotTableCacheDefinitions do not always have source ranges to update on row/column insertion/deletions.
Fix issue where deleting a column would only shift a PivotTable if at least one regular table existed in the worksheet.


EPPlus/ExcelWorksheet.cs
```diff
@@ -1415,8 +1415,10 @@ public void InsertRow(int rowFrom, int rows, int copyStylesFromRow)
 				}
 				foreach (var ptbl in PivotTables)
 				{
-					ptbl.Address = ptbl.Address.AddRow(rowFrom, rows);
-					ptbl.CacheDefinition.SourceRange.Address = ptbl.CacheDefinition.SourceRange.AddRow(rowFrom, rows).Address;
+					if (rowFrom <= ptbl.Address.End.Row)
+						ptbl.Address = ptbl.Address.AddRow(rowFrom, rows);
+					if (ptbl.CacheDefinition.CacheSource == eSourceType.Worksheet && rowFrom <= ptbl.CacheDefinition.SourceRange.End.Row)
```
emdelaney: Should we be checking that the source is referencing the right worksheet here? (Are we handling cross-sheet source references correctly?)

EPPlusTest/ExcelWorksheetTests.cs
```diff
@@ -3428,6 +3507,90 @@ public void InsertColumnThatWouldPushFormulasOffTheSheetThrowsException()
...
+
+		[TestMethod]
+		[DeploymentItem(@"..\..\Workbooks\PivotTableDataSourceTypeExternal.xlsx")]
+		public void DeleteColumnUpdatesPivotTableSourceRangeHandlesExternalDataSources()
+		{
+			var file = new FileInfo("PivotTableDataSourceTypeExternal.xlsx");
+			Assert.IsTrue(file.Exists);
+			using (var package = new ExcelPackage(file))
+			{
+				var worksheet = package.Workbook.Worksheets.First();
```
emdelaney: I'd be grateful if we could add a couple of unit tests that explicitly test the cross-sheet cases: for worksheet Pivot with a PivotTable that references data on worksheet Data:
-Inserting/deleting on Pivot does nothing
-Inserting/deleting on Data correctly updates the range.

EPPlusTest/ExcelAddressTest.cs
```diff
@@ -206,12 +206,13 @@ public void ChangeWorksheetNotApplied()
 		}
 		#endregion
 
-		#region AddRow Tests some other thing
```
emdelaney: Oops :D

PR #58, opened 2017-06-12
Bugs/801 support large function

NaKister:
Added test suites for the Large and Small Functions and provided support for these functions.


EPPlusTest/FormulaParsing/Excel/Functions/Math/LargeTests.cs
```diff
@@ -0,0 +1,262 @@
...
+				ws.Cells["A3"].Value = 2;
+				ws.Cells["A4"].Value = 789;
+				ws.Cells["A5"].Value = 3;
+				ws.Cells["B1"].Formula = "LARGE(A2:A5, 2)";
+				ws.Calculate();
+				Assert.AreEqual(45d, ws.Cells["B1"].Value);
+			}
+		}
+
+		[TestMethod]
```
emdelaney: If it's not an error case, we should also have some tests against mixed numbers, logical values, and text here (CreateArgs(27, true, "1000"), 1).

EPPlus/FormulaParsing/Excel/Functions/Math/Large.cs
```diff
@@ -40,24 +40,27 @@ public class Large : ExcelFunction
...
+			if (args.Value == null)
+				return new CompileResult(eErrorType.Num);
+			if (args.Value is string)
 				if (!ConvertUtil.TryParseNumericString(args.Value, out _))
 					if(!ConvertUtil.TryParseDateString(args.Value, out _))
 						return new CompileResult(eErrorType.Value);
 
 			var index = this.ArgToInt(arguments, 1) - 1;
-			var values = this.ArgsToDoubleEnumerable(new List<FunctionArgument> { args }, context);
+			var argsAsEnum = this.ArgsToObjectEnumerable(false, new List<FunctionArgument> { args }, context);
```
emdelaney: I might call this 'argumentValueList'.

EPPlus/FormulaParsing/Excel/Functions/Math/Large.cs
```diff
@@ -40,24 +40,27 @@ public class Large : ExcelFunction
...
+			var values = argsAsEnum.Where(arg => arg.GetType().IsPrimitive && false == arg is bool).Select(arg => arg);
```
emdelaney: I think the more common && (arg is bool == false)) is probably the more readable way to write this, looking on it again.

emdelaney: We can also get rid of the Select() call here since we don't need a different property (a common select pattern is `x => x.SomeProperty`, like `cats.select(cat => cat.Color)` to get a list of type `CatColor` instead of `Cat`, which we don't need here).

EPPlus/FormulaParsing/Excel/Functions/Math/Large.cs
```diff
@@ -40,24 +40,27 @@ public class Large : ExcelFunction
...
 			if (index < 0 || index >= values.Count())
 				return new CompileResult(eErrorType.Num);
 			var result = values.OrderByDescending(x => x).ElementAt(index);
-			return this.CreateResult(result, DataType.Decimal);
+			return this.CreateResult(result, DataType.Integer);
```
emdelaney: I think you want decimal here, right, since you can pass in a list of decimals and get a decimal out?

EPPlus/FormulaParsing/Excel/Functions/Math/Small.cs
```diff
@@ -55,11 +57,12 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
...
 
 			var index = this.ArgToInt(arguments, 1) - 1;
-			var values = this.ArgsToDoubleEnumerable(new List<FunctionArgument> { args }, context);
+			var argsAsEnum = this.ArgsToObjectEnumerable(false, new List<FunctionArgument> { args }, context);
+			var values = argsAsEnum.Where(arg => arg.GetType().IsPrimitive && false == arg is bool).Select(arg => arg);
 			if (index < 0 || index >= values.Count())
 				return new CompileResult(eErrorType.Num);
 			var result = values.OrderBy(x => x).ElementAt(index);
-			return this.CreateResult(result, DataType.Decimal);
+			return this.CreateResult(result, DataType.Integer);
```
emdelaney: Same question.

EPPlusTest/FormulaParsing/Excel/Functions/Math/SmallTests.cs
```diff
@@ -254,6 +270,32 @@ public void SmallWithInvalidArgumentReturnsPoundValue()
...
+			using (var package = new ExcelPackage())
+			{
+				var worksheet = package.Workbook.Worksheets.Add("Sheet1");
+				worksheet.Cells["B1"].Formula = "SMALL({TRUE, 67, \"1000\"}, 1)";
+				//worksheet.Cells["B2"].Formula = "SMALL({TRUE, 99, \"345\"}, 3)";
+				worksheet.Calculate();
+				Assert.AreEqual(67d, worksheet.Cells["B1"].Value);
+				//Assert.AreEqual(eErrorType.Num, ((ExcelErrorValue)worksheet.Cells["B2"].Value).Type);
+			}
+		}
```
emdelaney: We should add a test that has a list of doubles.

PR #59, opened 2017-06-12
Bugs/800 provide support for exp function

colewfrench:


EPPlus/FormulaParsing/Excel/Functions/Math/FactDouble.cs
```diff
@@ -0,0 +1,67 @@
...
+*******************************************************************************/
+using System.Collections.Generic;
+using System.Linq;
+using OfficeOpenXml.FormulaParsing.ExpressionGraph;
+using OfficeOpenXml.Utils;
+
+namespace OfficeOpenXml.FormulaParsing.Excel.Functions.Math
+{
+	/// <summary>
+	/// Returns the mathematical double factorial of a number (not the double data type).
```
emdelaney: What is "the mathematical double factorial of a number"? It looks like this is just a Factorial method that can take dates as well as numbers? If we can clean up / make this comment more precise that would be awesome :)

EPPlus/FormulaParsing/Excel/Functions/Math/FactDouble.cs
```diff
@@ -0,0 +1,67 @@
...
+		/// </summary>
+		/// <param name="arguments">Contains the number to perform the double factorial operation on.</param>
+		/// <param name="context">Unused in the method.</param>
+		/// <returns>Returns the double factorial of the given number, or an <see cref="ExcelErrorValue"/> if the input is invalid.</returns>
+		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
+		{
+			if (this.ValidateArguments(arguments, 1) == false)
+				return new CompileResult(eErrorType.Value);
+			if (arguments.ElementAt(0).ValueIsExcelError)
+				return new CompileResult(arguments.ElementAt(0).ValueAsExcelErrorValue);
```
emdelaney: I wonder if we should be doing this check more generally.

EPPlus/FormulaParsing/Excel/Functions/Math/Fact.cs
```diff
@@ -1,23 +1,67 @@
...
+		/// <param name="context">Unused in the method.</param>
+		/// <returns>Returns the factorial of the given number, or an <see cref="ExcelErrorValue"/> if the given input is invalid.</returns>
 		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
 		{
-			if (ValidateArguments(arguments, 1) == false)
+			if (this.ValidateArguments(arguments, 1) == false)
+				return new CompileResult(eErrorType.Value);
+			if (arguments.ElementAt(0).ValueIsExcelError)
+				return new CompileResult(arguments.ElementAt(0).ValueAsExcelErrorValue);
+			// Despite its name, this method just turns objects into numbers. Its original purpose was for parsing
```
emdelaney: I don't like this comment. We should either rename the method (TryParseObjectToDouble?), or just remove this comment entirely.

EPPlus/FormulaParsing/Excel/Functions/Math/FactDouble.cs
```diff
@@ -0,0 +1,67 @@
...
+			// Despite its name, this method just turns objects into numbers. Its original purpose was for parsing
+			// any object as if it represented a date, but it's basically the same as ArgToInt, except that it can
```
emdelaney: Same deal here, we should get rid of this comment one way or another.

EPPlus/FormulaParsing/Excel/Functions/Math/FactDouble.cs
emdelaney: I'd make a task to go back and get this right for our other functions. Also consider if we should write a method on the base class, like `TryGetErrorArgument(args, out excelErrorValue error)`, to catch and return the appropriate error code.

PR #61, opened 2017-06-13
Bugs/863 change validate arguments to catch error values

colewfrench:


EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -110,11 +110,41 @@ protected object GetFirstValue(IEnumerable<FunctionArgument> val)
...
+		///								if this method returns false. The default value is the value given for <paramref name="errorOnInvalidCount"/>.</param>
+		/// <param name="errorOnInvalidCount">The desired <see cref="eErrorType"/> to receive if this method returns false.</param>
+		/// <returns>Returns true if there are at least the <paramref name="minLength"/> number of arguments present 
+		///			 and none of the arguments contain an <see cref="ExcelErrorValue"/>, and returns false if otherwise.</returns>
+		protected bool ArgumentsAreValid(IEnumerable<FunctionArgument> arguments, int minLength, out eErrorType argumentError, eErrorType errorOnInvalidCount = eErrorType.Value)
+		{
+			argumentError = errorOnInvalidCount;
+			if (!this.ArgumentCountIsValid(arguments, minLength))
+				return false;
+			foreach (var argument in arguments)
```
emdelaney: The more Jet-style way to write this would be to use LINQ instead of a for-loop, something like:
`var error = arguments.FirstOrDefault(arg => arg.ValueIsExcelError);
if (error != null)
{
 argumentError = argument.ValueAsExcelErrorValue.Type;
 return false;
}`

PR #69, opened 2017-06-19
Adds functionality to set an ExcelRangeBase Formula without clearing its value.

EvanSchallerer:


EPPlus/ExcelRangeBase.cs
```diff
@@ -254,30 +254,26 @@ private static void Set_Value(ExcelRangeBase range, object value, int row, int c
 			if (sfi != null) range._worksheet._formulas.SetValue(row, col, string.Empty);
 			range._worksheet.SetValueInner(row, col, value);
 		}
-		private static void Set_Formula(ExcelRangeBase range, object value, int row, int col)
-		{
-			var f = range._worksheet._formulas.GetValue(row, col);
-			if (f is int && (int)f >= 0) range.UpdateSharedFormulaAnchors(range._worksheet.Cells[row, col]);
 
+		private static void Set_Formula(ExcelRangeBase range, object value, int row, int col, bool clearValue = true)
```
emdelaney: This is long-standing, but should be renamed to something more like SetFormula or SetFormulaInner or something.

EPPlus/ExcelRangeBase.cs
```diff
@@ -254,30 +254,26 @@ private static void Set_Value(ExcelRangeBase range, object value, int row, int c
...
 		}
-		private static void Set_Formula(ExcelRangeBase range, object value, int row, int col)
-		{
-			var f = range._worksheet._formulas.GetValue(row, col);
-			if (f is int && (int)f >= 0) range.UpdateSharedFormulaAnchors(range._worksheet.Cells[row, col]);
 
+		private static void Set_Formula(ExcelRangeBase range, object value, int row, int col, bool clearValue = true)
+		{
+			var formulaValue = range._worksheet._formulas.GetValue(row, col);
+			if (formulaValue is int && (int)formulaValue >= 0)
```
emdelaney: We could use the C#7 `if (formulaValue is int formulaInt && formulaInt >= 0)` syntax here, but I'm not sure it would be more readable.

EPPlus/ExcelRangeBase.cs
```diff
@@ -254,30 +254,26 @@ private static void Set_Value(ExcelRangeBase range, object value, int row, int c
...
+
 		/// <summary>
 		/// Handles shared formulas
 		/// </summary>
 		/// <param name="range">The range</param>
 		/// <param name="value">The  formula</param>
 		/// <param name="address">The address of the formula</param>
 		/// <param name="IsArray">If the forumla is an array formula.</param>
-		private static void Set_SharedFormula(ExcelRangeBase range, string value, ExcelAddress address, bool IsArray)
+		private static void Set_SharedFormula(ExcelRangeBase range, string value, ExcelAddress address, bool IsArray, bool clearValue = true)
```
emdelaney: This should be renamed to SetSharedFormula or SetSharedFormulaInner or something.

EPPlus/ExcelRangeBase.cs
```diff
@@ -286,7 +282,7 @@ private static void Set_SharedFormula(ExcelRangeBase range, string value, ExcelA
 			else if (address.Start.Row == address.End.Row && address.Start.Column == address.End.Column && !IsArray)             //is it really a shared formula? Arrayformulas can be one cell only
 			{
 				//Nope, single cell. Set the formula
```
emdelaney: This comment can be improved: `Single cells get individual formulas instead of shared formulas`, or something.

PR #73, opened 2017-06-20
Bugs/794 support sign c int abs functions

NaKister:
Changed the name of the CInt Class to IntFunction.


EPPlus/FormulaParsing/Excel/Functions/Math/Abs.cs
```diff
@@ -1,44 +1,74 @@
...
+		/// <returns>The absolute value of the argument given.</returns>
 		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
 		{
 			if (this.ArgumentsAreValid(arguments, 1, out eErrorType argumentError) == false)
 				return new CompileResult(argumentError);
-			var val = ArgToDecimal(arguments, 0);
-			if (val < 0)
+
+			var numberCandidate = arguments.ElementAt(0).Value;
+			if (numberCandidate is string)
```
emdelaney: Can we just replace all of this with ConvertUtil.TryParseObjectToOADate?

EPPlus/FormulaParsing/Excel/Functions/Math/Sign.cs
```diff
@@ -1,49 +1,79 @@
...
+			var numberCandidate = arguments.ElementAt(0).Value;
+			if (numberCandidate is string)
+			{
+				if (!ConvertUtil.TryParseNumericString(numberCandidate, out _))
+					if (!ConvertUtil.TryParseDateString(numberCandidate, out _))
+						if (!ConvertUtil.TryParseBooleanString(numberCandidate, out _))
+							return new CompileResult(eErrorType.Value);
+			}
+			else if (numberCandidate == null)
+			{
```
emdelaney: These brackets can go away.

EPPlus/FormulaParsing/Excel/Functions/Math/Sign.cs
```diff
@@ -1,49 +1,79 @@
...
+				return this.CreateResult(0d, DataType.Decimal);
+			}
+
+			var numberValue = this.ArgToDecimal(arguments, 0);
+			if (numberValue < 0)
 			{
```
emdelaney: These brackets can also go away.

EPPlus/FormulaParsing/Excel/Functions/Math/Sign.cs
```diff
@@ -1,49 +1,79 @@
...
 				result = -1;
 			}
-			else if (val > 0)
+			else if (numberValue > 0)
 			{
```
emdelaney: These, too.

EPPlus/FormulaParsing/Excel/Functions/Numeric/IntFunction.cs
```diff
@@ -0,0 +1,67 @@
...
+		/// <returns>The user specified argument as an integer value.</returns>
+		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
+		{
+			if (this.ArgumentsAreValid(arguments, 1, out eErrorType argumentError) == false)
+				return new CompileResult(argumentError);
+
+			var numberCandidate = arguments.ElementAt(0).Value;
+			if (numberCandidate is string)
+			{
+				if (!ConvertUtil.TryParseNumericString(numberCandidate, out _))
```
emdelaney: Same here. If these end up being necessary, I'd be tempted to do && and chain each item on its own line, instead of nesting them all in if statements.

EPPlus/FormulaParsing/Excel/Functions/Numeric/IntFunction.cs
```diff
@@ -0,0 +1,67 @@
+﻿/*******************************************************************************
```
emdelaney: Not a big deal for us right now, but going forward there's a git command (it's not integrated into VS, which is frustrating) to "move" files ( https://git-scm.com/docs/git-mv ) which tracks the source history better when files are moved or renamed that would be good to use :) It's tricky to do that retroactively though so we won't worry about it here.

EPPlus/FormulaParsing/Excel/Functions/Math/Abs.cs
```diff
@@ -51,23 +51,16 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
...
-						if (!ConvertUtil.TryParseBooleanString(numberCandidate, out _))
-							return new CompileResult(eErrorType.Value);
-			}
-			else if (numberCandidate == null)
-			{
+
+			if (numberCandidate == null)
 				return this.CreateResult(0d, DataType.Decimal);
-			}
+			if (!ConvertUtil.TryParseDateObjectToOADate(numberCandidate, out _))
```
emdelaney: Can we just use the out parameter here?

PR #75, opened 2017-06-22
Resolves issue when sqrefs are missing from conditionalFormattings, adds test.

EvanSchallerer:
* Discards conditional formattings and x14 rules that are missing sqref addresses
* Adds test
* Updates files to better meet conventions and standards


EPPlus/ConditionalFormatting/ExcelConditionalFormattingCollection.cs
```diff
@@ -67,20 +67,16 @@ public class ExcelConditionalFormattingCollection
...
-		/****************************************************************************************/
-
-		#region Private Properties
-		private List<IExcelConditionalFormattingRule> _rules = new List<IExcelConditionalFormattingRule>();
-		private ExcelWorksheet _worksheet = null;
-		#endregion Private Properties
-
-		/****************************************************************************************/
+		#region Class Variables
+		private List<IExcelConditionalFormattingRule> myRules = new List<IExcelConditionalFormattingRule>();
```
emdelaney: Let's consider making these readonly properties with getters instead of private class variables (readonly props can be initialized in the constructor).

EPPlusTest/ConditionalFormatting/ConditionalFormatting.cs
```diff
@@ -13,68 +13,35 @@ namespace EPPlusTest
...
-			{
-				return testContextInstance;
-			}
-			set
-			{
-				testContextInstance = value;
-			}
-		}
+		#region Class Variables
+		private static ExcelPackage myExcelPackage;
```
emdelaney: Let's make this a property.

EPPlusTest/ConditionalFormatting/ConditionalFormatting.cs
```diff
@@ -83,44 +50,34 @@ public void TwoColorScale()
...
 			var cf3 = ws.Cells["d12:D22,H12:H22"].ConditionalFormatting.AddFourIconSet(eExcelconditionalFormatting4IconsSetType.RedToBlack);
 			pck.SaveAs(new FileInfo(@"c:\temp\cf2.xlsx"));
 		}
-		/// <summary>
-		/// 
-		/// </summary>
+
 		[TestMethod]
 		[Ignore]
 		public void ReadConditionalFormattingError()
```
emdelaney: Any of these methods that have c:\temp\file.xlsx files can just be deleted, since we don't have their source files available.

PR #81, opened 2017-06-26
Bugs/794 add floor math and ceiling math functions

NaKister:
Added two new classes, FloorMath and CeilingMath which behave as FLOOR.MATH and CEILING.MATH, respectively.


EPPlus/FormulaParsing/Excel/Functions/Math/FloorMath.cs
```diff
@@ -0,0 +1,132 @@
...
+			bool exactChange;
+
+			if (this.ArgumentsAreValid(arguments, 1, out eErrorType argumentError) == false)
+				return new CompileResult(argumentError);
+
+			var numberCandidate = arguments.ElementAt(0).Value;
+			var significanceCandidate = arguments.ElementAt(1).Value;
+
+			if (numberCandidate == null)
+				return this.CreateResult(0d, DataType.Decimal);
```
emdelaney: The whitespace between all of these error cases can be removed.

EPPlus/FormulaParsing/Excel/Functions/Math/FloorMath.cs
```diff
@@ -0,0 +1,132 @@
...
+					divisionResult = number / significance;
+					multiple = (int)divisionResult;
+					exactChange = divisionResult == multiple;
+					if (exactChange)
+						return this.CreateResult(number, DataType.Decimal);
+					else if (significance > 0 && number < 0)
+						return this.CreateResult((multiple - 1) * significance, DataType.Decimal);
+					else
+						return this.CreateResult(multiple * significance, DataType.Decimal);
+					}
```
emdelaney: This bracket should be de-tabbed one.

EPPlus/FormulaParsing/Excel/Functions/Math/CeilingMath.cs
```diff
@@ -0,0 +1,151 @@
...
+{
+	/// <summary>
+	/// This class contains the formula for the CEILING.MATH Excel Function. 
+	/// </summary>
+	public class CeilingMath : ExcelFunction
+	{
+		/// <summary>
+		/// Takes the first user argument and rounds it up based on the optional second and third arguments.
+		/// </summary>
+		/// <param name="arguments">The user specified arguments.</param>
```
emdelaney: In here, we should specify what, exactly, the "user-specified arguments" are, and what they mean, so that we know that for example, "mode" is expected to be 0, 1, or -1. and decides whether or not the argument should be rounded up or down.

EPPlus/FormulaParsing/Excel/Functions/Math/CeilingMath.cs
```diff
@@ -0,0 +1,151 @@
...
+				if (exactChange)
+					return this.CreateResult(number, DataType.Decimal);
+				else if (significance > 0 && number < 0)
+					return this.CreateResult((multiple - 1) * significance, DataType.Decimal);
+				else if (number < 0)
+					return this.CreateResult((multiple + 1) * significance, DataType.Decimal);
+				else
+					return this.CreateResult(multiple * significance, DataType.Decimal);
+			}
+			else if (significance < 1 && significance > 0)
```
emdelaney: I wonder if there's a way to simplify these cases at all, there are a lot of them here. Once you have tests for all the cases, I'd look and see if it can be simplified at all, there are a lot of special cases that we might be able to consolidate.

EPPlus/FormulaParsing/Excel/Functions/Math/CeilingMath.cs
```diff
@@ -0,0 +1,151 @@
...
+				var modeCandidate = arguments.ElementAt(2).Value;
+				if (modeCandidate == null)
+				{
+					divisionResult = number / significance;
+					multiple = (int)divisionResult;
+					exactChange = divisionResult == multiple;
+					if (exactChange)
+						return this.CreateResult(number, DataType.Decimal);
+					else if (significance > 0 && number < 0)
+						return this.CreateResult((multiple + 1) * significance, DataType.Decimal);
```
emdelaney: I don't see a test for this case; for every case in the code here we should have at least one unit test. This will also help us refactor this code with confidence :)

EPPlus/FormulaParsing/Excel/Functions/Math/CeilingMath.cs
```diff
@@ -53,38 +55,23 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
...
 			if (significanceCandidate == null)
 				return this.CreateResult(System.Math.Ceiling(number), DataType.Decimal);
 			if (!ConvertUtil.TryParseDateObjectToOADate(significanceCandidate, out double significance))
 				return new CompileResult(eErrorType.Value);
 			if (significance < 0)
 				significance = -1 * significance;
-
 			if (arguments.Count() > 2)
 			{
 				var modeCandidate = arguments.ElementAt(2).Value;
```
emdelaney: We can do this cleverly with question marks: var modeCandidate = arguments.ElementAt(2)?.Value? ?? 0;

PR #82, opened 2017-06-27
Bugs/804 support rank function

NaKister:
Added three new test classes for the three Rank functions in Excel: RANK, RANK.EQ, and RANK.AVG. Made the Rank class more readable.

### DIFF
--- a/EPPlus/FormulaParsing/Excel/Functions/Math/Rank.cs
+++ b/EPPlus/FormulaParsing/Excel/Functions/Math/Rank.cs
@@ -6,23 +6,25 @@ using OfficeOpenXml.FormulaParsing.ExpressionGraph;
 namespace OfficeOpenXml.FormulaParsing.Excel.Functions.Math
 {
 	/// <summary>
-	/// This class contains the formulas for the RANK, RANK.EQ, and RANK.AVG Excel Functions. Based on what is passed into the 
-	/// constructor is the function that is executed (Specifically RANK and RANK.EQ v.s. RANK.AVG).
+	/// This class contains the formulas for the RANK, RANK.EQ, and RANK.AVG Excel Functions. The function to be executed is 
+	/// determined dynamically based on the arguments to the constructor (RANK and RANK.EQ vs. RANK.AVG).
 	/// </summary>
 	public class Rank : ExcelFunction
 	{
-		bool _isAvg;
+		private bool _isAvg;
 		/// <summary>
-		/// If the RANK.AVG Function is to be executed then true will be passed into the constructor. 
+		/// Creates a new RANK Function.
 		/// </summary>
-		/// <param name="isAvg"></param>
+		/// <param name="isAvg">Indicates if the function should use the RANK.AVG variant instad of the default RANK.EQ
+		/// implementation.</param>
 		public Rank(bool isAvg = false)
 		{
-			_isAvg = isAvg;
+			this._isAvg = isAvg;
 		}
 		/// <summary>
-		/// Takse the user specified arguments and returns the rank of the first argument; the position of the first argument
-		/// in relation to the other numbers in a list. 
+		/// Takse the user specified arguments and returns the  the position of the first argument in relation to the 
+		/// other numbers in a sorted list. For example in a scrambled list of the numbers 1-100 the number 99 would have 
+		/// a rank of 2. 
 		/// </summary>
 		/// <param name="arguments">The user specified arguments, the first being the number whos rank is to be found and
 		/// the second is the list of numbers as a cell reference.</param>
@@ -49,6 +51,7 @@ namespace OfficeOpenXml.FormulaParsing.Excel.Functions.Math
 			var numberList = new List<double>();
 			foreach (var cell in reference.ValueAsRangeInfo)
 			{
+				// Boolean and string values in a referenced range are ignored by the function. 
 				if (cell.Value is bool)
 					continue;
 				var valAsDouble = Utils.ConvertUtil.GetValueDouble(cell.Value, false, true);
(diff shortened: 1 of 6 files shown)


EPPlus/FormulaParsing/Excel/Functions/Math/Rank.cs
```diff
@@ -5,64 +5,86 @@
 
 namespace OfficeOpenXml.FormulaParsing.Excel.Functions.Math
 {
+	/// <summary>
+	/// This class contains the formulas for the RANK, RANK.EQ, and RANK.AVG Excel Functions. Based on what is passed into the 
+	/// constructor is the function that is executed (Specifically RANK and RANK.EQ v.s. RANK.AVG).
```
emdelaney: This sentence needs some grammar help, maybe ¨The function to be executed is determined dynamically based on the arguments to the constructor (RANK and RANK.EQ v.s. RANK.AVG).

EPPlus/FormulaParsing/Excel/Functions/Math/Rank.cs
```diff
@@ -5,64 +5,86 @@
 
 namespace OfficeOpenXml.FormulaParsing.Excel.Functions.Math
 {
+	/// <summary>
+	/// This class contains the formulas for the RANK, RANK.EQ, and RANK.AVG Excel Functions. Based on what is passed into the 
+	/// constructor is the function that is executed (Specifically RANK and RANK.EQ v.s. RANK.AVG).
+	/// </summary>
 	public class Rank : ExcelFunction
 	{
 		bool _isAvg;
```
emdelaney: This should be a private property (private bool IsAverage { get; }) in its own region.

EPPlus/FormulaParsing/Excel/Functions/Math/Rank.cs
```diff
@@ -5,64 +5,86 @@
...
 {
+	/// <summary>
+	/// This class contains the formulas for the RANK, RANK.EQ, and RANK.AVG Excel Functions. Based on what is passed into the 
+	/// constructor is the function that is executed (Specifically RANK and RANK.EQ v.s. RANK.AVG).
+	/// </summary>
 	public class Rank : ExcelFunction
 	{
 		bool _isAvg;
+		/// <summary>
+		/// If the RANK.AVG Function is to be executed then true will be passed into the constructor.
```
emdelaney: Creates a new RANK function.
Param comment: Indicates if the function should use the RANK.AVG variant instead of the default RANK.EQ implementation.

EPPlus/FormulaParsing/Excel/Functions/Math/Rank.cs
```diff
@@ -5,64 +5,86 @@
...
 	public class Rank : ExcelFunction
 	{
 		bool _isAvg;
+		/// <summary>
+		/// If the RANK.AVG Function is to be executed then true will be passed into the constructor. 
+		/// </summary>
+		/// <param name="isAvg"></param>
 		public Rank(bool isAvg = false)
 		{
 			_isAvg = isAvg;
```
emdelaney: this.IsAverage = ...

EPPlus/FormulaParsing/Excel/Functions/Math/Rank.cs
```diff
@@ -5,64 +5,86 @@
...
 		}
+		/// <summary>
+		/// Takse the user specified arguments and returns the rank of the first argument; the position of the first argument
```
emdelaney: I'd get rid of "the rank of the first argument;".

emdelaney: Is this the rank as in, in a scambled list of the numbers 1 to 100, the number 99 would have rank 2? Or it would have the index of its location returned, which could be any number between 1 and 100? This comment should explicitly indicate which of these is actually occurring : )

EPPlus/FormulaParsing/Excel/Functions/Math/Rank.cs
```diff
@@ -5,64 +5,86 @@
...
-			foreach (var c in refer.ValueAsRangeInfo)
+			var numberList = new List<double>();
+			foreach (var cell in reference.ValueAsRangeInfo)
 			{
-				var v = Utils.ConvertUtil.GetValueDouble(c.Value, false, true);
-				if (!double.IsNaN(v))
-				{
-					l.Add(v);
-				}
+				if (cell.Value is bool)
```
emdelaney: I'd maybe throw a comment in here noting that boolean and string values in a referenced range are ignored, since the string-values-are-ignored bit is not obvious from the code.

EPPlus/FormulaParsing/Excel/Functions/Math/Rank.cs
```diff
@@ -5,64 +5,86 @@
...
+			if (ascendingOrder)
 			{
-				ix = l.IndexOf(number) + 1;
+				rankResult = numberList.IndexOf(number) + 1;
 				if (_isAvg)
 				{
-					int st = Convert.ToInt32(ix);
-					while (l.Count > st && l[st] == number) st++;
-					if (st > ix) ix = ix + ((st - ix) / 2D);
+					int st = Convert.ToInt32(rankResult);
```
emdelaney: Let's use a better name than "st" :)

EPPlus/FormulaParsing/Excel/Functions/Math/Rank.cs
```diff
@@ -5,64 +5,86 @@
...
 			else
 			{
-				ix = l.LastIndexOf(number);
+				rankResult = numberList.LastIndexOf(number);
 				if (_isAvg)
 				{
-					int st = Convert.ToInt32(ix) - 1;
-					while (0 <= st && l[st] == number) st--;
-					if (st + 1 < ix) ix = ix - ((ix - st - 1) / 2D);
+					int intRank = Convert.ToInt32(rankResult) - 1;
```
emdelaney: Wow, this was pretty hard-to-read code before! Let's maybe rename intRank to duplicateResultCount or offset or something more specific / accurate?

EPPlusTest/FormulaParsing/Excel/Functions/Math/RankEqTests.cs
```diff
@@ -0,0 +1,394 @@
...
+			{
+				var worksheet = package.Workbook.Worksheets.Add("Sheet1");
+				worksheet.Cells["B1"].Formula = "DATE(2017, 5, 5)";
+				worksheet.Cells["B2"].Formula = "DATE(2017, 6, 15)";
+				worksheet.Cells["B3"].Formula = "RANK.EQ(42901, B1:B2)";
+				worksheet.Calculate();
+				Assert.AreEqual(1d, worksheet.Cells["B3"].Value);
+			}
+		}
+
```
emdelaney: We should throw a test in here like the test in the previous class where all the values are the same, to demonstrate how this function differs from the AVG variant.

PR #84, opened 2017-06-28
Bugs/807 stdev and var

NAccuardi:
Stdev.p and Stdev.s have both been implemented and tested. Getting this up here so it is not one giant pull request at the end of the task. Im going to push things as I get them done. Ill make sure to let you know as I do pushes.


EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,88 @@
...
+		/// The standard deviation is a measure of how widely values are dispersed from the average value (the mean).
+		/// Logical values and text representations of numbers that you type directly into the list of arguments are counted.
+		/// If an argument is an array or reference, only numbers in that array or reference are counted.Empty cells, logical values, text, or error values in the array or reference are ignored.
+		/// </summary>
+		/// <param name="arguments">Up too 254 individual arguments.</param>
+		/// <param name="context">Unused, this is information about where the function is being executed.</param>
+		/// <returns>The standard deviation based on a sample.</returns>
+		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
+		{
+			var args = ArgsToDoubleEnumerable(IgnoreHiddenValues, false, arguments, context);
```
emdelaney: this.IgnoreHidden...

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using OfficeOpenXml.FormulaParsing.ExpressionGraph;
+using OfficeOpenXml.Utils;
+using MathObj = System.Math;
+
+namespace OfficeOpenXml.FormulaParsing.Excel.Functions.Math
+{
+	/// <summary>
+	/// Estimates standard deviation based on a sample.
```
emdelaney: This is an awesome comment. Clear, concise, accurate, descriptive, with terms that can be googled if necessary.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,88 @@
...
+		/// <param name="context">Unused, this is information about where the function is being executed.</param>
+		/// <returns>The standard deviation based on a sample.</returns>
+		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
+		{
+			var args = ArgsToDoubleEnumerable(IgnoreHiddenValues, false, arguments, context);
+			foreach (var item in arguments)
+			{
+				if (item.IsExcelRange)
+				{
+					if (item.ValueFirst is double || item.ValueFirst is int)
```
emdelaney: Same question here, I think we can hack this with a range that has the first value invalid but later values valid (or, alternately, a range that has the first value valid and later invalid values).

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,88 @@
...
+						continue;
+					return new CompileResult(eErrorType.Div0);
+				}
+			}
+			if (!TryStandardDeviationOnASamplePopulation(args, arguments, out double StanderedDeviation))
```
emdelaney: this.

emdelaney: standardDeviation

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,88 @@
...
+				return new CompileResult(eErrorType.Value);
+			return CreateResult(StanderedDeviation, DataType.Decimal);
```
emdelaney: this.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		private static bool TryStandardDeviationOnASamplePopulation(IEnumerable<double> values, IEnumerable<FunctionArgument> arguments, out double StanderedDeviation)
```
emdelaney: standardDeviation

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,88 @@
...
+		{
+			List<double> listOfValues = new List<double>();
+			foreach (var item in values)
+			{
+				StanderedDeviation = 0.0;
```
emdelaney: this initialization should go outside the loop.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,88 @@
...
+			if (StanderedDeviation == 0 && listOfValues.All(x => x == listOfValues.First()))
+				return false;
+			return true;
+		}
+
+		private static double Var(IEnumerable<double> args)
+		{
+			double avg = args.Average();
+			double d = args.Aggregate(0.0, (total, next) => total += System.Math.Pow(next - avg, 2));
+			return (d/ (args.Count() - 1));
```
emdelaney: d /

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,88 @@
...
+			foreach (var item in values)
+			{
+				StanderedDeviation = 0.0;
+				var checkThis = ConvertUtil.TryParseDateObjectToOADate(item, out double result12);
+				if (!ConvertUtil.TryParseDateObjectToOADate(item, out double result))
+					return false;
+				listOfValues.Add(result);
+			}
+
+			StanderedDeviation = MathObj.Sqrt(Var(listOfValues));
```
emdelaney: this.Var?

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,88 @@
...
+			if (StanderedDeviation == 0 && listOfValues.All(x => x == listOfValues.First()))
```
emdelaney: Can we do this check somewhere else / better? It smells kind of weird to have it here.

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevPTests.cs
```diff
@@ -0,0 +1,332 @@
...
+using OfficeOpenXml.FormulaParsing.Excel.Functions.Math;
+
+namespace EPPlusTest.FormulaParsing.Excel.Functions.Math
+{
+
+	[TestClass]
+	public class StdevPTests : MathFunctionsTestBase
+	{
+		#region StdevP Function(Execute) Tests
+
```
emdelaney: This extra whitespace can be removed.

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevPTests.cs
```diff
@@ -0,0 +1,332 @@
...
+				Assert.AreEqual(30.835045, (double)worksheet.Cells["A11"].Value, .00001);
+				Assert.AreEqual(22.32011425, (double)worksheet.Cells["A12"].Value, .00001);
+			}
+		}
+
+		[TestMethod]
+		public void StdevpIsGivenNumbersAsInputs()
+		{
+			var function = new StdevP();
+			var result1 = function.Execute(FunctionsHelper.CreateArgs(-1,0,1), this.ParsingContext);
```
emdelaney: (-1, 0, 1)

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevPTests.cs
```diff
@@ -0,0 +1,332 @@
...
+				worksheet.Cells["B1"].Value = "5.5";
+				worksheet.Cells["B2"].Value = "6.6";
+				worksheet.Cells["B3"].Value = "7.7";
+				worksheet.Cells["A4"].Formula = "=stdev.p(A1,A2,A3)";
+				worksheet.Cells["B4"].Formula = "=stdev.p(B1,B2,B3)";
+				worksheet.Calculate();
+				Assert.AreEqual(eErrorType.Div0, ((ExcelErrorValue)worksheet.Cells["A4"].Value).Type);
+				Assert.AreEqual(eErrorType.Div0, ((ExcelErrorValue)worksheet.Cells["B4"].Value).Type);
+			}
+		}
```
emdelaney: We should add a bunch of test cases with references to mixed cell value types-- empty, int, parseable string, unparseable string, etc (see the logic comments above for the specific cases that should be tested).

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevSTests.cs
```diff
@@ -0,0 +1,332 @@
...
+				worksheet.Cells["B2"].Value = "6.6";
+				worksheet.Cells["B3"].Value = "7.7";
+				worksheet.Cells["A4"].Formula = "=stdev.s(A1,A2,A3)";
+				worksheet.Cells["B4"].Formula = "=stdev.s(B1,B2,B3)";
+				worksheet.Calculate();
+				Assert.AreEqual(eErrorType.Div0, ((ExcelErrorValue)worksheet.Cells["A4"].Value).Type);
+				Assert.AreEqual(eErrorType.Div0, ((ExcelErrorValue)worksheet.Cells["B4"].Value).Type);
+			}
+		}
+		#endregion
```
emdelaney: Same here, we should add tests that cover the logic you have around range arguments like B2:B4.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -25,22 +25,56 @@
...
-		private static double StandardDeviation(IEnumerable<double> values)
+		private static bool TryStandardDeviationEntirePopulation(IEnumerable<double> values, IEnumerable<FunctionArgument> arguments, out double StanderedDeviation)
 		{
-			double avg = values.Average();
-			return MathObj.Sqrt(values.Average(v => MathObj.Pow(v - avg, 2)));
+			List<double> listOfValues = new List<double>();
+			foreach (var item in values)
+			{
+				StanderedDeviation = 0.0;
+				var checkThis = ConvertUtil.TryParseDateObjectToOADate(item, out double result12);
```
NAccuardi: Oops. Some of this was me figuring out functionality. Ill clean it up.

EPPlus/EPPlus.csproj
```diff
@@ -584,6 +585,7 @@
     <Compile Include="FormulaParsing\ExpressionGraph\FunctionCompilers\IfFunctionCompiler.cs" />
     <Compile Include="FormulaParsing\ExpressionGraph\FunctionCompilers\IfNaFunctionCompiler.cs" />
     <Compile Include="FormulaParsing\ExpressionGraph\FunctionCompilers\LookupFunctionCompiler.cs" />
+    <Compile Include="FormulaParsing\ExpressionGraph\FunctionCompilers\ResolveCellReferencesAsRangeFunctionCompiler.cs" />
```
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -25,22 +25,56 @@
...
 		{
 			var args = ArgsToDoubleEnumerable(IgnoreHiddenValues, false, arguments, context);
-			return CreateResult(StandardDeviation(args), DataType.Decimal);
+			foreach (var item in arguments)
+			{
+				if (item.IsExcelRange)
+				{
+					if(item.ValueFirst is double || item.ValueFirst is int)
+						continue;
+					return new CompileResult(eErrorType.Div0);
```
emdelaney: Can you pass in an excelRange here that has an empty first value, but valid second and/or third values? This seems likely to break in such a case.

NAccuardi: Good idea. Ill make a test case for it.

EPPlus/FormulaParsing/ExpressionGraph/FunctionCompilers/ResolveCellReferencesAsRangeFunctionCompiler.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+	class ResolveCellReferencesAsRangeFunctionCompiler : DefaultCompiler
+	{
+		public ResolveCellReferencesAsRangeFunctionCompiler(ExcelFunction function) : base(function) { }
+
+		public override CompileResult Compile(IEnumerable<Expression> children, ParsingContext context)
+		{
+			// EPPlus handles operators as members of the child expression instead of as functions of their own.
+			// We want to exclude those children whose values are part of an operator expression (since they will be resolved by the operator).
+			var ignoreOperators = children.Where(child => child.Children.All(grandkid => grandkid.Operator == null));
```
emdelaney: This will probably not do what you want it to do-- you probably want to do childrenToResolveAsRange = children.Where(child -> child is ExcelAddressExpression), and get rid of the rest of the LINQ. Consult with Matt/Cole/Hughes/Zach as appropriate here (the debugger in a function where you care about this, with a breakpoint, will let us make sure we're grabbing the right nodes in the AST).

EPPlus/FormulaParsing/Excel/Functions/DoubleEnumerableArgConverter.cs
```diff
@@ -53,6 +53,11 @@ public virtual IEnumerable<double> ConvertArgs(bool ignoreHidden, bool ignoreErr
 						 {
 							 argList.Add(ConvertUtil.GetValueDouble(arg.Value));
 						 }
+						 if(arg.Value is string)
+						 {
+							 ConvertUtil.TryParseDateObjectToOADate(arg.Value, out double result);
```
NAccuardi: It has to be set up like this to account for all the test cases so far. We do want to add the result if it fails. That is how it deals with things like "string" or "another string".

NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -25,22 +25,56 @@
...
+	/// Calculates standard deviation based on the entire population given as arguments (ignores logical values and text).
+	/// </summary>
 	public class StdevP : HiddenValuesHandlingFunction
 	{
+		/// <summary>
+		/// The standard deviation is a measure of how widely values are dispersed from the average value (the mean).
+		/// Logical values and text representations of numbers that you type directly into the list of arguments are counted.
+		/// If an argument is an array or reference, only numbers in that array or reference are counted.Empty cells, logical values, text, or error values in the array or reference are ignored.
+		/// </summary>
+		/// <param name="arguments">Up too 254 individual arguments.</param>
```
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -25,22 +25,56 @@
...
+					return new CompileResult(eErrorType.Div0);
+				}
+			}
+			if(!TryStandardDeviationEntirePopulation(args, arguments, out double StanderedDeviation))
+				return new CompileResult(eErrorType.Value);
+			return CreateResult(StanderedDeviation, DataType.Decimal);
 		}
 
-		private static double StandardDeviation(IEnumerable<double> values)
+		private static bool TryStandardDeviationEntirePopulation(IEnumerable<double> values, IEnumerable<FunctionArgument> arguments, out double StanderedDeviation)
```
NAccuardi: More remnants from my trying different things. ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: I got rid of the unused tryparse.

NAccuardi: Looks like you were right. I didn't realize we could use values like a list, that is going to clean this portion up a lot.

NAccuardi: That made that so much cleaner. :) ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: Is the 'this.' for the TryStandardDeviationOnASamplePopulation?

emdelaney: Yep!

NAccuardi: Disregard. Got it fixed.

NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: Cleaned it up like in Stdev.P. It is a lot simpler now. ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: ✔️

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevPTests.cs
NAccuardi: ✔️

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevPTests.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: I think the reason it smells weird is because it is replicating weird behavior in excel. It has to do with how excel is dealing with when strings eg, "string", are passed into Stdev.S. I think the is the best place for it, but I would be more then willing to move it somewhere else if needed.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: You were absolutely right about that breaking. Working it now.

NAccuardi: Fixed. Just needed to add another check in case it is an empty cell. ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: This handles ranges that start with a valid or invalid input, as well as ranges that contain invalid ranges, and ranges that are only valid ranges. ✔️

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevPTests.cs
NAccuardi: Do you mean a test case kind of like =STDEV.S(1, TRUE, "6/17/2011 2:00", "02:00 am")?

emdelaney: Yeah. We should do that directly, and then also do the same thing with a range reference, like  STDEV.S(B2:B10) where those cells have a sweet mix of data types.

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevSTests.cs
NAccuardi: I have added a large number of tests to handle ranges. ✔️

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevPTests.cs
NAccuardi: I have added these tests. For STDEV.S The mixed range reference test fails. For STDEV.P if it is a mixed test by cell refrence or range is fails. working on this now.

EPPlus/FormulaParsing/Excel/Functions/DoubleEnumerableArgConverter.cs
```diff
@@ -40,19 +40,33 @@ public virtual IEnumerable<double> ConvertArgs(bool ignoreHidden, bool ignoreErr
 						 foreach (var cell in arg.ValueAsRangeInfo)
 						 {
 							 if (!ignoreErrors && cell.IsExcelError) throw new ExcelErrorValueException(ExcelErrorValue.Parse(cell.Value.ToString()));
-							 if (!CellStateHelper.ShouldIgnore(ignoreHidden, cell, context) && ConvertUtil.IsNumeric(cell.Value))
+							 if (!CellStateHelper.ShouldIgnore(ignoreHidden, cell, context) && ConvertUtil.IsNumeric(cell.Value) && !Boolean.TryParse(cell.Value.ToString() , out bool result))
 							 {
-								 argList.Add(cell.ValueDouble);
+								 argList.Add(cell.ValueDouble );
```
emdelaney: There's an extra space here.

EPPlus/FormulaParsing/Excel/Functions/DoubleEnumerableArgConverter.cs
```diff
@@ -40,19 +40,33 @@ public virtual IEnumerable<double> ConvertArgs(bool ignoreHidden, bool ignoreErr
 						 foreach (var cell in arg.ValueAsRangeInfo)
 						 {
 							 if (!ignoreErrors && cell.IsExcelError) throw new ExcelErrorValueException(ExcelErrorValue.Parse(cell.Value.ToString()));
-							 if (!CellStateHelper.ShouldIgnore(ignoreHidden, cell, context) && ConvertUtil.IsNumeric(cell.Value))
+							 if (!CellStateHelper.ShouldIgnore(ignoreHidden, cell, context) && ConvertUtil.IsNumeric(cell.Value) && !Boolean.TryParse(cell.Value.ToString() , out bool result))
 							 {
-								 argList.Add(cell.ValueDouble);
+								 argList.Add(cell.ValueDouble );
+							 }
+							 if (!CellStateHelper.ShouldIgnore(ignoreHidden, cell, context) && ConvertUtil.IsNumeric(cell.Value) && Boolean.TryParse(cell.Value.ToString(), out bool result2))
```
emdelaney: We can avoid making some method calls twice here by caching the result of each of these conditions the first time we check them and doing some intelligent nesting.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -25,22 +25,62 @@
...
+		/// If an argument is an array or reference, only numbers in that array or reference are counted.Empty cells, logical values, text, or error values in the array or reference are ignored.
+		/// </summary>
+		/// <param name="arguments">Up to 254 individual arguments.</param>
+		/// <param name="context">Unused, this is information about where the function is being executed.</param>
+		/// <returns>The standard deviation based on the entire population.</returns>
 		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
 		{
-			var args = ArgsToDoubleEnumerable(IgnoreHiddenValues, false, arguments, context);
-			return CreateResult(StandardDeviation(args), DataType.Decimal);
+			//NOTE: This follows the Functionality of excel which is diffrent from the excel documentation.
```
emdelaney: I'd maybe not all-caps this, so //Note: this follows...

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -25,22 +25,62 @@
...
+			//If you pass in a null Stdev.P(1,1,1,,) it will treat those emtpy spaces as zeros insted of ignoring them.
+			List<double> listToDoStandardDeviationOn = new List<double>();
+			var args = ArgsToDoubleEnumerable(this.IgnoreHiddenValues, false, arguments, context);
```
emdelaney: this.argsToDouble...

emdelaney: Since we're not iterating over this imnmediately, let's move it down below this first for-loop.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -25,22 +25,62 @@
...
+			foreach (var item in arguments)
```
emdelaney: I wonder if this should be done in ARgsToDoubleEnumerable, instead of here? Does this apply generally to all args-to-double cases?

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -25,22 +25,62 @@
...
+			return this.CreateResult(standardDeviation, DataType.Decimal);
 		}
 
-		private static double StandardDeviation(IEnumerable<double> values)
+		private bool TryStandardDeviationEntirePopulation(List<double> listToDoStandardDeviationOn, out double standardDeviation)
 		{
-			double avg = values.Average();
-			return MathObj.Sqrt(values.Average(v => MathObj.Pow(v - avg, 2)));
+			standardDeviation = MathObj.Sqrt(this.VarPopulation(listToDoStandardDeviationOn));
+			if (standardDeviation == 0 && listToDoStandardDeviationOn.All(x => x == -1))
```
emdelaney: 1. This seems really weird. Why are we checking against -1 here?
2. This seems like it will break spectacularly if you pass in an empty population (like, say, an unparseable string).

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -25,22 +25,62 @@
...
+				return false;
+			return true;
+		}
+
+		private double VarPopulation(List<double> listOfDoubles)
+		{
+			double avg = listOfDoubles.Average();
+			double d = listOfDoubles.Aggregate(0.0, (total, next) => total += System.Math.Pow(next - avg, 2));
+			return (d / (listOfDoubles.Count()));
```
emdelaney: Specifically, this is where it will break-- if listOfDoubles has length zero.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,90 @@
...
+			//NOTE: This follows the Functionality of excel which is diffrent from the excel documentation.
+			//If you pass in a null Stdev.S(1,1,1,,) it will treat those emtpy spaces as zeros insted of ignoring them.
+			List<double> listToDoStandardDeviationOn = new List<double>();
+			var args = ArgsToDoubleEnumerable(this.IgnoreHiddenValues, false, arguments, context);
+			foreach (var item in arguments)
+			{
+
+				if (item.IsExcelRange)
+				{
+
```
emdelaney: There are some extra newlines here. I'd also lowercase the NOTE above. Same question about ArgsToDoubleEnumerable.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,90 @@
...
+					if (item.ValueFirst is double || item.ValueFirst is int || item.ValueFirst == null)
+					{
+						continue;
+					}
+					return new CompileResult(eErrorType.Div0);
+				}
+				if (item.ValueFirst == null)
+					listToDoStandardDeviationOn.Add(0.0);
+			}
+			foreach(var item in args)
```
emdelaney: We should add a space after this foreach ( but before the (.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,90 @@
...
+				listToDoStandardDeviationOn.Add(item);
+			var IfStanderedDeviationWorked = this.TryStandardDeviationOnASamplePopulation(listToDoStandardDeviationOn, out double standardDeviation);
```
emdelaney: I'd maybe call this 'standardDeviationSucceeded' or ''worked or similar.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,90 @@
...
+			if (!IfStanderedDeviationWorked)
+				return new CompileResult(eErrorType.Value);
+			return this.CreateResult(standardDeviation, DataType.Decimal);
+		}
+
+		private bool TryStandardDeviationOnASamplePopulation(List<double> listToDoStandardDeviationOn, out double standardDeviation)
+		{
+			standardDeviation = MathObj.Sqrt(this.VarSampleSize(listToDoStandardDeviationOn));
+			if (standardDeviation == 0 && listToDoStandardDeviationOn.All(x => x == -1))
```
emdelaney: Same question here about how weird it is to check against -1.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
```diff
@@ -0,0 +1,90 @@
...
+				return false;
+			return true;
+		}
+
+		private double VarSampleSize(List<double> listOfDoubles)
+		{
+			double avg = listOfDoubles.Average();
+			double d = listOfDoubles.Aggregate(0.0, (total, next) => total += System.Math.Pow(next - avg, 2));
+			return (d / (listOfDoubles.Count() - 1));
```
emdelaney: This will break on both zero and one-length double lists.

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
```diff
@@ -0,0 +1,109 @@
...
+		/// The standard deviation is a measure of how widely values are dispersed from the average value (the mean).
+		/// Arguments can be the following: numbers; names, arrays, or references that contain numbers; text representations of numbers; or logical values, such as TRUE and FALSE, in a reference.
+		/// If an argument is an array or reference, only values in that array or reference are used. Empty cells and text values in the array or reference are ignored.
+		/// </summary>
+		/// <param name="arguments">Up too 254 individual arguments.</param>
+		/// <param name="context">Unused, this is information about where the function is being executed.</param>
+		/// <returns>The standard deviation based on a sample.</returns>
+		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
+		{
+			//NOTE: This follows the Functionality of excel which is diffrent from the excel documentation.
```
emdelaney: this can be lower-cased.

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
```diff
@@ -0,0 +1,109 @@
...
+{
+	/// <summary>
+	/// Estimates standard deviation based on a sample (includes logical values and text in the sample).
+	/// </summary>
+	public class Stdeva : HiddenValuesHandlingFunction
+	{
+		/// <summary>
+		/// The standard deviation is a measure of how widely values are dispersed from the average value (the mean).
+		/// Arguments can be the following: numbers; names, arrays, or references that contain numbers; text representations of numbers; or logical values, such as TRUE and FALSE, in a reference.
+		/// If an argument is an array or reference, only values in that array or reference are used. Empty cells and text values in the array or reference are ignored.
```
emdelaney: This is a good header comment :)

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
```diff
@@ -0,0 +1,109 @@
...
+		/// </summary>
+		/// <param name="arguments">Up too 254 individual arguments.</param>
+		/// <param name="context">Unused, this is information about where the function is being executed.</param>
+		/// <returns>The standard deviation based on a sample.</returns>
+		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
+		{
+			//NOTE: This follows the Functionality of excel which is diffrent from the excel documentation.
+			//If you pass in a null Stdev.S(1,1,1,,) it will treat those emtpy spaces as zeros insted of ignoring them.
```
emdelaney: this should be stdevA instead of stdev.S? Probably a copy-paste error in a bunch of these files.

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
```diff
@@ -0,0 +1,109 @@
...
+			List<double> listToDoStandardDeviationOn = new List<double>();
+			bool DontAddBoolsTwice = false;
```
emdelaney: maybe call this skipAddingBooleanValue or booleanAlreadyIncluded? It's generally good software design practice to name variables, especially boolean variables, 'positively'. Makes them easier to read and maintain.

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
```diff
@@ -0,0 +1,109 @@
...
+						if (cell.Value is string)
+							listToDoStandardDeviationOn.Add(0.0);
+					}
+					if (item.ValueFirst is double || item.ValueFirst is int || item.ValueFirst == null)
+						continue;
+				}
+				if (item.ValueFirst == null)
+					listToDoStandardDeviationOn.Add(0.0);
+				if (item.Value is ExcelDataProvider.IRangeInfo itemRange)
+				{
```
emdelaney: I see we're doing something slightly different here, so we probably will need either a boolean flag or a different ArgsToDoubleEnumerable-type call to handle it. I'd be tempted to move all of this logic into a call resembling ArgsToDoubleEnumerable, I think it will be much cleaner.

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
```diff
@@ -0,0 +1,109 @@
...
+					listToDoStandardDeviationOn.Add(item);
+			if (!this.TryStandardDeviationEntireSamplePopulation(listToDoStandardDeviationOn, out double standardDeviation))
+				return new CompileResult(eErrorType.Value);
+			return this.CreateResult(standardDeviation, DataType.Decimal);
+		}
+
+		private bool TryStandardDeviationEntireSamplePopulation(List<double> listToDoStandardDeviationOn, out double standardDeviation)
+		{
+			standardDeviation = MathObj.Sqrt(this.VarSamplePopulation(listToDoStandardDeviationOn));
+			if (standardDeviation == 0 && listToDoStandardDeviationOn.All(x => x == -1))
```
emdelaney: Same question about this weirdness.

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
```diff
@@ -0,0 +1,109 @@
...
+				return false;
+			return true;
+		}
+
+		private double VarSamplePopulation(List<double> listOfDoubles)
+		{
+			double avg = listOfDoubles.Average();
+			double d = listOfDoubles.Aggregate(0.0, (total, next) => total += System.Math.Pow(next - avg, 2));
+			return (d / (listOfDoubles.Count() - 1));
```
emdelaney: This will also break on a population of size 0 or 1.

EPPlus/FormulaParsing/Excel/Functions/Math/Stdevpa.cs
```diff
@@ -0,0 +1,108 @@
...
+		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
+		{
+			//NOTE: This follows the Functionality of excel which is diffrent from the excel documentation.
+			//If you pass in a null Stdev.S(1,1,1,,) it will treat those emtpy spaces as zeros insted of ignoring them.
+			List<double> listToDoStandardDeviationOn = new List<double>();
+			var args = ArgsToDoubleEnumerable(this.IgnoreHiddenValues, false, arguments, context);
+			foreach (var item in arguments)
+			{
+
+
```
emdelaney: There's nothing specific to this file, but all of the comments from the previous files still apply to this one (whitespace, stdev.S comment, extracting this code to ArgsToDoubleEnumerable, this., small-range-failures, etc).

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
```diff
@@ -23,18 +23,59 @@
...
+		/// Variance measures how far a data set is spread out.
+		/// Logical values, and text representations of numbers that you type directly into the list of arguments are counted.
+		/// If an argument is an array or reference, only numbers in that array or reference are counted. Empty cells, logical values, text, or error values in the array or reference are ignored.
+		/// </summary>
+		/// <param name="arguments">Up too 254 individual arguments.</param>
+		/// <param name="context">Unused, this is information about where the function is being executed.</param>
+		/// <returns>The variance based on an entire population.</returns>
 		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
 		{
-			if (this.ArgumentsAreValid(arguments, 1, out eErrorType argumentError) == false)
```
emdelaney: Should we be validating these functions' arguments at all before we try to do anything? Seems like it might be valuable.

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
```diff
@@ -23,18 +23,59 @@
...
+			//If you pass in a null Stdev.S(1,1,1,,) it will treat those emtpy spaces as zeros insted of ignoring them.
+			List<double> listToDoVarianceOn = new List<double>();
+			var args = ArgsToDoubleEnumerable(this.IgnoreHiddenValues, false, arguments, context);
+			foreach (var item in arguments)
+			{
+				if (item.IsExcelRange)
+				{
+					if (item.ValueFirst is double || item.ValueFirst is int || item.ValueFirst == null)
+						continue;
+					return new CompileResult(eErrorType.Div0);
```
emdelaney: this seems like the wrong place to return this, generally-- we probably want to improve the ArgsToDoubleEnumerable call, and then if the length of the resulting enumerable is zero, we should return the div0 error.

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
```diff
@@ -23,18 +23,59 @@
...
+				else if (item.ValueFirst == null)
+				{
+					listToDoVarianceOn.Add(0.0);
+				}
+			}
+			foreach (var item in args)
+				listToDoVarianceOn.Add(item);
+			if(!this.TryVarPopulation(listToDoVarianceOn, out double VarPopulation))
+				return new CompileResult(eErrorType.Value);
+
```
emdelaney: Some extra whitespace here.

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
```diff
@@ -23,18 +23,59 @@
...
+			foreach (var item in arguments)
+			{
+				if (item.IsExcelRange)
+				{
+					if (item.ValueFirst is double || item.ValueFirst is int || item.ValueFirst == null)
+						continue;
+					return new CompileResult(eErrorType.Div0);
+				}
+				else if (item.ValueFirst == null)
+				{
```
emdelaney: Our coding standard wants these brackets to go away.

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
```diff
@@ -23,18 +23,59 @@
...
+					listToDoVarianceOn.Add(0.0);
+				}
+			}
+			foreach (var item in args)
+				listToDoVarianceOn.Add(item);
```
emdelaney: But it also wants brackets here, because for-loops always get brackets (the bracket rules are pretty arcane).

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
```diff
@@ -23,18 +23,59 @@
...
+			if(!this.TryVarPopulation(listToDoVarianceOn, out double VarPopulation))
+				return new CompileResult(eErrorType.Value);
+
+			return new CompileResult(VarPopulation, DataType.Decimal);
+		}
+
+		private bool TryVarPopulation(List<double> listOfDoubles, out double VarPopulation)
```
emdelaney: I'd call this populationVariance, or just variance.

EPPlus/FormulaParsing/Excel/Functions/Math/VarS.cs
```diff
@@ -0,0 +1,81 @@
...
+		/// Logical values, and text representations of numbers that you type directly into the list of arguments are counted.
+		/// If an argument is an array or reference, only numbers in that array or reference are counted. Empty cells, logical values, text, or error values in the array or reference are ignored.
+		/// </summary>
+		/// <param name="arguments">Up too 254 individual arguments.</param>
+		/// <param name="context">Unused, this is information about where the function is being executed.</param>
+		/// <returns>The variance based on a sample.</returns>
+		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
+		{
+			//NOTE: This follows the Functionality of excel which is diffrent from the excel documentation.
+			//If you pass in a null Stdev.S(1,1,1,,) it will treat those emtpy spaces as zeros insted of ignoring them.
```
emdelaney: All the same comments here.

EPPlus/FormulaParsing/Excel/Functions/Math/Vara.cs
```diff
@@ -0,0 +1,102 @@
...
+							continue;
+						}
+					}
+					return this.CreateResult(0d, DataType.Decimal);
+				}
+			}
+			foreach (var item in args)
+				if (!DontAddBoolsTwice)
+					listToDoVarianceOn.Add(item);
+			if (!this.TryVarSample(listToDoVarianceOn, out double VarSample))
```
emdelaney: And here, too.

EPPlus/FormulaParsing/Excel/Functions/Math/Varpa.cs
```diff
@@ -0,0 +1,100 @@
...
+				}
+			}
+			foreach (var item in args)
+				listToDoVarianceOn.Add(item);
+			if (!this.TryVarPopulation(listToDoVarianceOn, out double VarPopulation))
+				return new CompileResult(eErrorType.Value);
+			return new CompileResult(VarPopulation, DataType.Decimal);
+		}
+
+		private bool TryVarPopulation(List<double> listOfDoubles, out double VarPopulation)
```
emdelaney: and also here.

EPPlus/FormulaParsing/ExpressionGraph/FunctionCompilers/ResolveCellReferencesAsRangeCompiler.cs
```diff
@@ -0,0 +1,54 @@
...
+*
+* For code change notes, see the source control history.
+*******************************************************************************/
+using System.Collections.Generic;
+using System.Linq;
+using OfficeOpenXml.FormulaParsing.Excel.Functions;
+
+namespace OfficeOpenXml.FormulaParsing.ExpressionGraph.FunctionCompilers
+{
+	class ResolveCellReferencesAsRangeCompiler : DefaultCompiler
```
emdelaney: Let's do one more code review here to address these changes, and then you'll want to merge back from master to your feature branch, since there's an existing compiler there that we should use here :) Yay fun merge conflicts.

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevPTests.cs
```diff
@@ -0,0 +1,773 @@
...
+				worksheet.Cells["B2"].Value = "6.6";
+				worksheet.Cells["B3"].Value = "7.7";
+				worksheet.Cells["A4"].Formula = "=Stdev.p(A1:A3)";
+				worksheet.Cells["B4"].Formula = "=Stdev.p(B1:B3)";
+				worksheet.Calculate();
+				Assert.AreEqual(eErrorType.Div0, ((ExcelErrorValue)worksheet.Cells["A4"].Value).Type);
+				Assert.AreEqual(eErrorType.Div0, ((ExcelErrorValue)worksheet.Cells["B4"].Value).Type);
+			}
+		}
+		#endregion
```
emdelaney: We should add a bunch of boundary-case tests for each of these: too few arguments, empty first argument, etc.

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevSTests.cs
```diff
@@ -0,0 +1,773 @@
...
+				worksheet.Cells["B4"].Value = "12:00";
+				worksheet.Cells["B5"].Value = 32;
+				worksheet.Cells["B6"].Value = 76;
+
+				worksheet.Cells["B7"].Value = 2;
+				worksheet.Cells["B8"].Value = 3;
+				worksheet.Cells["B9"].Value = 5;
+				worksheet.Cells["B10"].Value = 7;
+				worksheet.Cells["B11"].Value = 45;
+
```
emdelaney: Lots of extra whitespace in these that should be removed eventually.

EPPlus/FormulaParsing/Excel/Functions/DoubleEnumerableArgConverter.cs
NAccuardi: I think I got all of them cleaned up here. ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
NAccuardi: :) ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
NAccuardi: Yep! That is absolutely what happened.

NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
NAccuardi: The funny part is I had brackets here until I went and "cleaned up" :P

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/VarS.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/Vara.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/Varpa.cs
NAccuardi: ✔️

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevSTests.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: 1. The -1 is what gets returned from ArgstodoubleEnumerable. What happens is from there it drills down into TryParseDateObjectToOADate. If TryParseDateObjectToOADate is not able to return a proper OADate it returns -1.

This -1 is only happens when the string does not contain a date and is a string like "a string" or "another string"

2. I will have to check this. Do you have an example of an unparasable string I can use to build a test case?

emdelaney: Hmm. For (1) we should probably be using the return value of TryParseObjectToOADate instead of adding the returned value since there's no guarantee about the out parameter for Try-do methods that return false.

emdelaney: For (2) I was thinking of the case you identified with "string" or "Cats" or similar.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: Fixed this in all the new functions. ✔️

NAccuardi: Is there a reason to move it? Or just good practice?

emdelaney: It makes it easier to read. It's a little weird, to me, that we'd create a collection of args and then iterate over a different collection, arguments. Could be easily overlooked and mis-interpreted by a future developer.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: This is specifically for the stdev and var methods. There might be a better way to handle it though. The issue I ran into was args was a IEnumerable<double>  so i wasn't able to add items to it. The way I got around it was to take the elements out of the IEnumerable and make a list with them that would allow me to add numbers to it. 

This needed to happen specifically because of how excel is handling the empty inputs.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: For (2) I have a test case that is ("string", "another string", "a third string") Ill add another test case that just looks at a single "string" input.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
NAccuardi: ✔️ That does make it alot easier to wrap your head around.

EPPlus/FormulaParsing/Excel/Functions/Math/VarP.cs
NAccuardi: Good call. ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: Good call. In my head I was thinking of it like declaring an variable at the top. I put them in a spot that makes more sense. ✔️

EPPlus/FormulaParsing/Excel/Functions/DoubleEnumerableArgConverter.cs
NAccuardi: I think I was able to follow your intent. We shall find out when I push again. :P ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevS.cs
NAccuardi: Fixed this. ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/Stdeva.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/DoubleEnumerableArgConverter.cs
```diff
@@ -34,41 +34,46 @@ public class DoubleEnumerableArgConverter : CollectionFlattener<double>
...
-						if (!ignoreErrors && arg.ValueIsExcelError) throw new ExcelErrorValueException(arg.ValueAsExcelErrorValue);
-						if (ConvertUtil.IsNumeric(arg.Value) && !CellStateHelper.ShouldIgnore(ignoreHidden, arg, context))
+			{
+				if (arg.IsExcelRange)
+				{
+					foreach (var cell in arg.ValueAsRangeInfo)
+					{
+						var shouldIgnore = CellStateHelper.ShouldIgnore(ignoreHidden, cell, context);
+						var isNumeric = ConvertUtil.IsNumeric(cell.Value);
+						var isABoolean = Boolean.TryParse(cell.Value.ToString(), out bool result2);
```
emdelaney: Should we be checking the type of the value (cell.Value is bool) instead here? I'm wondering of "true" should come back as a boolean value or not.

EPPlus/FormulaParsing/Excel/Functions/DoubleEnumerableArgConverter.cs
```diff
@@ -34,41 +34,46 @@ public class DoubleEnumerableArgConverter : CollectionFlattener<double>
...
+
+						if (!ignoreErrors && cell.IsExcelError) throw new ExcelErrorValueException(ExcelErrorValue.Parse(cell.Value.ToString()));
```
emdelaney: Let's put the throw on its own line :)

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -49,23 +50,40 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
...
 			{
-				if (item.IsExcelRange)
+				if(item.ValueAsRangeInfo != null)
 				{
-					if (item.ValueFirst is double || item.ValueFirst is int || item.ValueFirst == null )
-						continue;
-					return new CompileResult(eErrorType.Div0);
+					foreach (var cell in item.ValueAsRangeInfo)
+					{
+						if (TryToParseValuesFromInputArgumentByRefrenceOrRange(this.IgnoreHiddenValues, false, cell, context, out double numberToAddToList, out bool onlyStringInputsGiven1))
```
emdelaney: this.TryParseValuesFrom...

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -49,23 +50,40 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
...
+							listToDoStandardDeviationOn.Add(numberToAddToList);
+						onlyStringInputsGiven = onlyStringInputsGiven1;
+						if(cell.Value is bool)
+							return new CompileResult(eErrorType.Div0);
+					}
+				}
+				else
+				{
+
```
emdelaney: There's some extra whitespace here.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -49,23 +50,40 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
...
-					if (item.ValueFirst is double || item.ValueFirst is int || item.ValueFirst == null )
-						continue;
-					return new CompileResult(eErrorType.Div0);
+					foreach (var cell in item.ValueAsRangeInfo)
+					{
+						if (TryToParseValuesFromInputArgumentByRefrenceOrRange(this.IgnoreHiddenValues, false, cell, context, out double numberToAddToList, out bool onlyStringInputsGiven1))
+							listToDoStandardDeviationOn.Add(numberToAddToList);
+						onlyStringInputsGiven = onlyStringInputsGiven1;
+						if(cell.Value is bool)
+							return new CompileResult(eErrorType.Div0);
```
emdelaney: This smells weird, since there might be other valid values in the list of arguments to work on, right?

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -49,23 +50,40 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
...
+				if(item.ValueAsRangeInfo != null)
 				{
-					if (item.ValueFirst is double || item.ValueFirst is int || item.ValueFirst == null )
-						continue;
-					return new CompileResult(eErrorType.Div0);
+					foreach (var cell in item.ValueAsRangeInfo)
+					{
+						if (TryToParseValuesFromInputArgumentByRefrenceOrRange(this.IgnoreHiddenValues, false, cell, context, out double numberToAddToList, out bool onlyStringInputsGiven1))
+							listToDoStandardDeviationOn.Add(numberToAddToList);
+						onlyStringInputsGiven = onlyStringInputsGiven1;
```
emdelaney: I think you can pass this in directly as `out onlyStringInputsGiven`.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -49,23 +50,40 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
...
+						if(cell.Value is bool)
+							return new CompileResult(eErrorType.Div0);
+					}
+				}
+				else
+				{
+
+					if (TryToParseValuesFromInputArgument(this.IgnoreHiddenValues, false, item, context, out double numberToAddToList, out bool onlyStringInputsGiven2))
+						listToDoStandardDeviationOn.Add(numberToAddToList);
+					onlyStringInputsGiven = onlyStringInputsGiven2;
```
emdelaney: Same here, we can just set this via the out parameter directly. Do we want to `and` it instead? I'd guess that's what we want based on how it's being used below in a mixed string- and numeric-cell-references test case.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -49,23 +50,40 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
...
-			foreach (var item in args)
-			{
-				listToDoStandardDeviationOn.Add(item);
 			}
+
+			if(onlyStringInputsGiven)
+				return new CompileResult(eErrorType.Value);
+
+			if (listToDoStandardDeviationOn.Count() == 0)
+				return new CompileResult(eErrorType.Div0);
```
emdelaney: This should be the only place the Div0 error gets returned in these methods.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -85,6 +103,54 @@ private double VarPopulation(List<double> listOfDoubles)
 			double d = listOfDoubles.Aggregate(0.0, (total, next) => total += System.Math.Pow(next - avg, 2));
 			return (d / (listOfDoubles.Count()));
 		}
+
+		private bool TryToParseValuesFromInputArgumentByRefrenceOrRange(bool ignoreHidden, bool ignoreErrors, ExcelDataProvider.ICellInfo valueToCheck, ParsingContext context, out double numberToWorkWith, out bool onlyStrings)
```
emdelaney: Let's call these TryParse... or TryGet..., move them to the base class eventually (make them `protected`), and either make them static or use this._context or whatever instead of passing it in as a parameter.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -85,6 +103,54 @@ private double VarPopulation(List<double> listOfDoubles)
...
+
+		private bool TryToParseValuesFromInputArgumentByRefrenceOrRange(bool ignoreHidden, bool ignoreErrors, ExcelDataProvider.ICellInfo valueToCheck, ParsingContext context, out double numberToWorkWith, out bool onlyStrings)
+		{
+			var shouldIgnore = CellStateHelper.ShouldIgnore(ignoreHidden, valueToCheck, context);
+			var isNumeric = ConvertUtil.IsNumeric(valueToCheck.Value);
+			var isABoolean = Boolean.TryParse(valueToCheck.Value.ToString(), out bool result2);
+
+			if (!shouldIgnore && isNumeric && !isABoolean)
+			{
+				numberToWorkWith = valueToCheck.ValueDouble;
```
emdelaney: Let's call this out parameter something like "parsedValue".

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
```diff
@@ -85,6 +103,54 @@ private double VarPopulation(List<double> listOfDoubles)
...
+				onlyStrings = false;
```
emdelaney: similarly, let's make this a little more self-documenting by calling this something like 'valueWasParsedFromString'.

emdelaney: For the other variables in this function, we should either make them more self-descriptive (ignoreHiddenWhat, exactly?), and/or add a header comment. I'd be very tempted, personally, to make this a public static method and test it independently.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
emdelaney: And maybe be more precise, like TryGetNumericValueFromRange().

EPPlus/FormulaParsing/Excel/Functions/DoubleEnumerableArgConverter.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/DoubleEnumerableArgConverter.cs
NAccuardi: ✔️ Good call.

EPPlusTest/FormulaParsing/Excel/Functions/Math/StdevPTests.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: So the basic form of this was from a public method. I took it form doubleEnumerableConverter and mad it purpose built for this method. I thought this would be best since each one of the stdev.p, stdev.s, stdeva, stdevpa, var.p, var.s, vara, and varpa all have slightly different ways of handling each one. 

Insted of making a single thing fit all the cases, I thought it would be better to make one for each.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: So I have been struggling with this a bit. The two bools catch two different edge sets. I would love to get together today and Ill show you what I mean. I am sure you will instantly see a better way of handling this. 

Is returning a type of error only once just a good coding convention? I assume because it would be confusing for other coders down the timeline to follow what was going on if they needed to do any work in here.

NAccuardi: Looks like that might of taken care of it! ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevP.cs
NAccuardi: I renamed and cleaned this up a bit. Should make a bit more sense at first glance. ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevAndVarHelperClass.cs
```diff
@@ -0,0 +1,217 @@
...
+using System.Linq;
+using OfficeOpenXml.Utils;
+using MathObj = System.Math;
+
+namespace OfficeOpenXml.FormulaParsing.Excel.Functions.Math
+{
+	/// <summary>
+	/// This is a helper class that is used for all standard deviation and variance methods.
+	/// </summary>
+	class StdevAndVarHelperClass
```
emdelaney: Minor nitpick, but we hate to see the word "and" in names around here, so we should probably rename this to "StatisticsFunctionHelper".

EPPlus/FormulaParsing/Excel/Functions/Math/StdevAndVarHelperClass.cs
```diff
@@ -0,0 +1,217 @@
...
+	{
+		/// <summary>
```
emdelaney: These guys should all get a #region.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevAndVarHelperClass.cs
```diff
@@ -0,0 +1,217 @@
...
+		///  This is used by the Stdev and Var functions to parse the values it receives from a cell range or refreence.
+		/// </summary>
+		/// <param name="IgnoreHiddenValues">This controls wether you should skip Hidden Values or not.</param>
+		/// <param name="valueToParse">This is the value that will be parsed.</param>
+		/// <param name="context">Unused, this is information about where the function is being executed.</param>
+		/// <param name="includeLogicals">A flag to control wether you should parse logicals or not.</param>
+		/// <param name="parsedValue">This out value is the parse value in double form.</param>
+		/// <param name="theInputContainedOnlyStrings">This bool flag shows if only string inputs were given.</param>
+		/// <returns>Returns true if the parseing succeded and false if it fails.</returns>
+		public static bool TryToParseValuesFromInputArgumentByRefrenceOrRange(bool IgnoreHiddenValues, ExcelDataProvider.ICellInfo valueToParse, ParsingContext context, bool includeLogicals, out double parsedValue, out bool theInputContainedOnlyStrings)
```
emdelaney: Now that these are all public, they should get their own unit tests if possible :)

EPPlus/FormulaParsing/Excel/Functions/Math/StdevAndVarHelperClass.cs
```diff
@@ -0,0 +1,217 @@
...
+					theInputContainedOnlyStrings = false;
+					return true;
+				}
+				parsedValue = 0.0;
+				theInputContainedOnlyStrings = false;
+				return false;
+			}
+			else
+			{
+				var shouldIgnore = CellStateHelper.ShouldIgnore(IgnoreHiddenValues, valueToParse, context);
```
emdelaney: There's a lot of duplicate code here that we should un-duplicate by putting it before the if check.

emdelaney: this.IgnoreHiddenValues

EPPlus/FormulaParsing/Excel/Functions/Math/StdevAndVarHelperClass.cs
```diff
@@ -0,0 +1,217 @@
...
+		/// </summary>
+		/// <param name="args">A list of inputs to have their variance calculated.</param>
+		/// <returns>The Variance for a Sample population</returns>
+		public static double VarianceForASample(List<double> args)
+		{
+			double avg = args.Average();
+			double d = args.Aggregate(0.0, (total, next) => total += System.Math.Pow(next - avg, 2));
+			return (d / (args.Count() - 1));
+		}
+		/// <summary>
```
emdelaney: There should be newlines between each of these methods and the next method's header comments :)

EPPlus/FormulaParsing/Excel/Functions/Math/VarMethods.cs
```diff
@@ -24,18 +24,13 @@
...
 using System.Linq;
-using OfficeOpenXml.FormulaParsing.Exceptions;
 
 namespace OfficeOpenXml.FormulaParsing.Excel.Functions.Math
 {
 	internal static class VarMethods
 	{
 		private static double Divide(double left, double right)
 		{
-			if (System.Math.Abs(right - 0d) < double.Epsilon)
```
emdelaney: Is there a reason we're getting rid of this? This seems like it was probably a nuanced bugfix, to me, so if we don't need to get rid of it we probably should leave it around for now.

EPPlus/FormulaParsing/Excel/Functions/Math/StdevAndVarHelperClass.cs
NAccuardi: minor nitpicks make me write better code so I welcome them! ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevAndVarHelperClass.cs
NAccuardi: Unduped as much of the code as I could. ✔️

NAccuardi: I can't so that since they are static. Any suggestions?

EPPlus/FormulaParsing/Excel/Functions/Math/StdevAndVarHelperClass.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevAndVarHelperClass.cs
NAccuardi: ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevAndVarHelperClass.cs
emdelaney: Oh, <className>.

EPPlus/FormulaParsing/Excel/Functions/Math/VarMethods.cs
NAccuardi: I actually meant to get rid of that entire class. I moved all of these too the StatsHelper class. I think this was causing an issue. I can add it to the statshelper class and do some work to see if I can get it working in there.

NAccuardi: This class get to lie into the future per our discussion. ✔️

EPPlus/FormulaParsing/Excel/Functions/Math/StdevAndVarHelperClass.cs
NAccuardi: Like StatisticsFunctionHelper.IgnoreHiddenValues?

NAccuardi: never mind that didn't work either.

emdelaney: Yeah, or if it's actually a local (function-scope) variable, it should be lower-cased.

PR #85, opened 2017-06-29
Fixes ExpressionConverter invalid cast when converting to doubles.

EvanSchallerer:


EPPlus/FormulaParsing/ExpressionGraph/ExpressionConverter.cs
```diff
@@ -32,16 +32,46 @@
...
+	/// Represents a converter used to convert values to expressions.
+	/// </summary>
 	public class ExpressionConverter : IExpressionConverter
 	{
+		#region Class Variables
+		private static IExpressionConverter myInstance;
+		#endregion
+
+		#region Properties
+		public static IExpressionConverter Instance
```
ZacharyPF: Comment public members.

EPPlusTest/FormulaParsing/ExpressionGraph/ExpressionConverterTests.cs
```diff
@@ -8,19 +9,24 @@ namespace EPPlusTest.FormulaParsing.ExpressionGraph
 	[TestClass]
 	public class ExpressionConverterTests
 	{
-		private IExpressionConverter _converter;
+		#region Class Variables
+		private IExpressionConverter myConverter;
```
ZacharyPF: Just make this a property.
We only use "my" for backing class variables and those are only referenced within the property they back.

EvanSchallerer: I was thinking of just removing it all together, it doesn't appear to buy anything over instantiating it in each test method.

PR #88, opened 2017-07-01
Fixed issue where ExcelComment.Text variable was not populated after copying from an ExcelRangeBase.

maxackley:


EPPlus/ExcelComment.cs
```diff
@@ -41,11 +41,10 @@ namespace OfficeOpenXml
 	public class ExcelComment : ExcelVmlDrawingComment
 	{
 		internal XmlHelper _commentHelper;
```
emdelaney: Let's make this a getter-only property, and give it a #region to live in :)

PR #92, opened 2017-07-03
Bugs/982 sum if sum ifs functions

NaKister:
Added support for the SumIf and SumIfs functions. 
Also changed the layout of the testing package in MathFunctionsTestBase so it can be used for future testing.


EPPlus/FormulaParsing/Excel/Functions/Math/SumIf.cs
```diff
@@ -45,6 +45,7 @@ public class SumIf : HiddenValuesHandlingFunction
 	{
 		private readonly ExpressionEvaluator _evaluator;
 		#region <ExcelFunction> Overrides
+		// These two constructors can probably be combined into one. Because of this there are not comments for them.
```
emdelaney: Let's remove this comment and add some actual comments :)

EPPlus/FormulaParsing/Excel/Functions/Math/SumIf.cs
```diff
@@ -45,6 +45,7 @@ public class SumIf : HiddenValuesHandlingFunction
 	{
 		private readonly ExpressionEvaluator _evaluator;
```
emdelaney: Is this still used? If not, we should get rid of it :)

EPPlus/FormulaParsing/Excel/Functions/Math/SumIf.cs
```diff
@@ -58,102 +59,104 @@ public SumIf(ExpressionEvaluator evaluator)
...
+			var cellRangeToCheck = arguments.ElementAt(0).Value as ExcelDataProvider.IRangeInfo;
+			if (cellRangeToCheck == null)
+				return new CompileResult(eErrorType.Value);
+			string criteriaString = null;
+			if (arguments.ElementAt(1).Value is ExcelDataProvider.IRangeInfo criteriaRange)
 			{
-				var sumRange = arguments.ElementAt(2).Value as ExcelDataProvider.IRangeInfo;
-				retVal = CalculateWithSumRange(args, criteria, sumRange, context);
+				if (criteriaRange.IsMulti)
+					return new CompileResult(0d, DataType.Decimal);
```
emdelaney: Is this really always zero? If C1:C3 are 1, 2, and 3, and we do SUMIF(C1:C3, C1:C3), I get one and not zero.

NaKister: Could you send me the workbook where you got this result or how you set this up because I'm performing this same operation and I'm getting 0.

NaKister:
Fixed a bug in the FormulaParser that caused it to parse cell ranges incorrectly. I also took that solution and implemented it into the IfHelper class so that the SumIf, SumIfs, AverageIf, and AverageIfs would be updated to reflect the cell range behavior in Excel.

EPPlus/FormulaParsing/FormulaParser.cs
```diff
@@ -153,6 +152,59 @@ internal virtual object ParseCell(IEnumerable<Token> tokens, string worksheet, i
 								 row, column);
 							_parsingContext.Configuration.Logger.Log(_parsingContext, msg);
 						}
+
```
emdelaney: There's some extra whitespace here that can be removed.

EPPlus/FormulaParsing/FormulaParser.cs
```diff
@@ -153,6 +152,59 @@ internal virtual object ParseCell(IEnumerable<Token> tokens, string worksheet, i
...
+							var endRow = rangeInfo.Address.End.Row;
+
+							var startCol = rangeInfo.Address.Start.Column;
+							var endCol = rangeInfo.Address.End.Column;
+
+							if(startCol != endCol)
+								return ExcelErrorValue.Create(eErrorType.Value);
+
+							if (rangeAddressRow == startRow)
+							{
```
emdelaney: All of these one-line if checks can have their brackets removed.

EPPlus/FormulaParsing/Excel/Functions/Math/IfHelper.cs
```diff
@@ -200,5 +202,95 @@ public static bool IsNumeric(object numericCandidate, bool excludeBool = false)
...
+		/// Takes a cell range and coverts it into a single value criteria. 
+		/// </summary>
+		/// <param name="arguments">The cell range that will be condensed into a single value.</param>
+		/// <param name="context">The context the function is being called in.</param>
+		/// <returns>A single value criteria as an integer.</returns>
+		public static int CalculateCriteria(IEnumerable<FunctionArgument> arguments, ParsingContext context)
+		{
+			var criteriaCandidate = arguments.ElementAt(1).ValueAsRangeInfo.Address;
+			var functionLocation = context.Scopes.Current.Address;
+			var worksheet = context.ExcelDataProvider.GetRange("Sheet1", 1, 1, "A1").Worksheet;
```
emdelaney: This seems unlikely to work in general, since the sheetname might not be "Sheet1".

EPPlus/FormulaParsing/Excel/Functions/Math/IfHelper.cs
```diff
@@ -200,5 +202,95 @@ public static bool IsNumeric(object numericCandidate, bool excludeBool = false)
...
 				numericCandidate is TimeSpan);
 		}
+	
+		/// <summary>
+		/// Takes a cell range and coverts it into a single value criteria. 
+		/// </summary>
+		/// <param name="arguments">The cell range that will be condensed into a single value.</param>
+		/// <param name="context">The context the function is being called in.</param>
+		/// <returns>A single value criteria as an integer.</returns>
+		public static int CalculateCriteria(IEnumerable<FunctionArgument> arguments, ParsingContext context)
```
emdelaney: This should be independently tested with its own unit tests (which might help us make this more generalizable). It might make sense to figure out which context properties you need to do the operation (the worksheet and the current location?), and pass those in to this method instead of passing the entire parsing context.

EPPlus/FormulaParsing/Excel/Functions/Math/IfHelper.cs
```diff
@@ -202,22 +202,23 @@ public static bool IsNumeric(object numericCandidate, bool excludeBool = false)
...
-		/// <param name="context">The context the function is being called in.</param>
+		/// <param name="arguments">The cell range that will be condensed inot a single value.</param>
+		/// <param name="worksheet">The current worksheet that is being used.</param>
+		/// <param name="rowLocation">The row location of the cell that is calling this function.</param>
+		/// <param name="colLocation">The column location of the cell that is calling this function.</param>
 		/// <returns>A single value criteria as an integer.</returns>
-		public static int CalculateCriteria(IEnumerable<FunctionArgument> arguments, ParsingContext context)
+		public static int CalculateCriteria(IEnumerable<FunctionArgument> arguments, ExcelWorksheet worksheet, int rowLocation, int colLocation)
 		{
 			var criteriaCandidate = arguments.ElementAt(1).ValueAsRangeInfo.Address;
```
emdelaney: We should do some error / null-checking here.

EPPlus/FormulaParsing/Excel/Functions/Math/IfHelper.cs
```diff
@@ -202,22 +202,23 @@ public static bool IsNumeric(object numericCandidate, bool excludeBool = false)
...
-	
+
+
 		/// <summary>
-		/// Takes a cell range and coverts it into a single value criteria. 
+		/// Takes a cell range and converts it into a single value criteria
 		/// </summary>
-		/// <param name="arguments">The cell range that will be condensed into a single value.</param>
-		/// <param name="context">The context the function is being called in.</param>
+		/// <param name="arguments">The cell range that will be condensed inot a single value.</param>
```
emdelaney: into

EPPlusTest/FormulaParsing/Excel/Functions/Math/IfHelperTests.cs
```diff
@@ -0,0 +1,118 @@
...
+using static OfficeOpenXml.FormulaParsing.ExcelDataProvider;
+using OfficeOpenXml.FormulaParsing;
+using OfficeOpenXml.FormulaParsing.ExcelUtilities;
+
+namespace EPPlusTest.FormulaParsing.Excel.Functions.Math
+{
+	[TestClass]
+	public class IfHelperTests : MathFunctionsTestBase 
+	{
+		[TestMethod]
```
emdelaney: Let's stick these guys in a #region.

PR #93, opened 2017-07-05
DeleteRow and DeleteColumn hardening.

emdelaney:


EPPlus/ExcelWorksheet.cs
```diff
@@ -1801,9 +1804,12 @@ public void DeleteColumn(int columnFrom, int columns)
 							ptbl.CacheDefinition.SourceRange?.Worksheet == this &&
 							ptbl.CacheDefinition.SourceRange.IsName == false &&
 							columnFrom <= ptbl.CacheDefinition.SourceRange.End.Column)
-							ptbl.CacheDefinition.SourceRange.Address = ptbl.CacheDefinition.SourceRange.DeleteColumn(columnFrom, columns).Address;
+							ptbl.CacheDefinition.SourceRange.Address = ptbl.CacheDefinition.SourceRange.DeleteColumn(columnFrom, columns)?.Address ?? ExcelErrorValue.Values.Ref;
 					}
 				}
+				this.UpdateCharts(0, -columns, 0, columnFrom);
```
emdelaney: Correct, they should, in theory, be totally independent. I haven't seen anything in any of the files or xml that suggests interdependency such that the order of update calls would matter.

PR #101, opened 2017-08-02
Bugs/1143 enter design mode removes pictures

colewfrench:


EPPlus/Drawing/ExcelDrawings.cs
```diff
@@ -528,36 +476,33 @@ public void Dispose()
...
-			this._drawings.Remove(draw);
 		}
 
 		internal void ClearDrawings()
 		{
-			while (Count > 0)
+			while (this.Count > 0)
 			{
-				this.RemoveDrawing(0);
+				this.RemoveDrawing(this._drawings[0]);
```
emdelaney: I don't know if there's any weirdness in how these all update when they get shifted down, but this approach seems weird to me, I'd expect to see a  `for (int i = this.Count - 1, i >= 0; i--)` loop here naively instead of a `while`. We can probably leave it alone for now though.

EPPlus/Drawing/ExcelDrawings.cs
```diff
@@ -641,22 +583,22 @@ internal void AdjustHeight(int[,] pos)
...
+		///	</returns>
 		public ExcelDrawing this[string Name]
 		{
 			get
 			{
-				if (this._drawingNames.ContainsKey(Name))
-				{
-					return this._drawings[this._drawingNames[Name]];
-				}
+				var drawingsWithGivenName = this._drawings.Where((ExcelDrawing drawing) => drawing.Name.Equals(Name));
```
emdelaney: I believe the FirstOrDefault(lambda function) LINQ method will get the first drawing that meets the condition, or null if none do :)

emdelaney:
Oh, man, how did I miss that. Good catch!

EPPlusTest/DrawingTest.cs
```diff
@@ -709,6 +710,82 @@ public void DrawingRowheightDynamic()
 		}
 		#endregion
 
+		#region Additional Drawings Tests
+		[TestMethod]
+		public void GetDrawingByNameGetsCorrectDrawing()
+		{
+			var worksheet = _pck.Workbook.Worksheets.Add("sheet");
```
emdelaney: All of these _pck references should probably be `this._pck` or `base.`

EPPlusTest/DrawingTest.cs
```diff
@@ -709,6 +710,82 @@ public void DrawingRowheightDynamic()
...
+
+			Assert.AreEqual(3, worksheet.Drawings.Count);
+			ExcelDrawing retrievedDrawing = worksheet.Drawings["LineChart"];
+			Assert.AreEqual(lineChartDrawing, retrievedDrawing);
+			retrievedDrawing = worksheet.Drawings["Hexagon"];
+			Assert.AreEqual(hexagonDrawing, retrievedDrawing);
+			retrievedDrawing = worksheet.Drawings["Picture"];
+			Assert.AreEqual(pictureDrawing, retrievedDrawing);
+			Assert.AreEqual(3, worksheet.Drawings.Count);
+			retrievedDrawing = worksheet.Drawings["NonExistent Drawing"];
```
emdelaney: It would be good to have a test here that explicitly tests a workbook that has no drawings where we try to access a nonexistent drawing.

EPPlusTest/DrawingTest.cs
```diff
@@ -709,6 +710,82 @@ public void DrawingRowheightDynamic()
...
+
+		[TestMethod]
+		public void DrawingWithMultipleDrawingsWithTheSameName()
+		{
+			var worksheet = _pck.Workbook.Worksheets.Add("sheet");
+			var picture1 = worksheet.Drawings.AddPicture("drawing", Resources.Test1);
+			var picture2 = worksheet.Drawings.AddPicture("drawing", Resources.Test1);
+			var shape1 = worksheet.Drawings.AddShape("drawing", eShapeStyle.Hexagon);
+			var shape2 = worksheet.Drawings.AddShape("drawing", eShapeStyle.Hexagon);
+			Assert.AreEqual(picture1, worksheet.Drawings["drawing"]);
```
emdelaney: I'd maybe throw a comment in here noting that for multiple drawings with the same name, the first matching drawing in the Drawings collection is returned.

EPPlusTest/DrawingTest.cs
```diff
@@ -709,6 +710,82 @@ public void DrawingRowheightDynamic()
...
+		[TestMethod]
+		public void ClearDrawings()
+		{
+			var worksheet = _pck.Workbook.Worksheets.Add("sheet");
+			var lineChartDrawing = worksheet.Drawings.AddChart("LineChart", eChartType.Line);
+			var hexagonDrawing = worksheet.Drawings.AddShape("Hexagon", eShapeStyle.Hexagon);
+			var pictureDrawing = worksheet.Drawings.AddPicture("Picture", Resources.Test1);
+			Assert.AreEqual(3, worksheet.Drawings.Count);
+			worksheet.Drawings.ClearDrawings();
+			Assert.AreEqual(0, worksheet.Drawings.Count);
```
emdelaney: We should save the workbook and try to re-open it somewhere, that might be better suited to Jet-side regression tests or to ExcelWorksheet/workbook/package tests but we should make sure that we can take a saved workbook with zero drawings; add drawings to it, save it; open it and find the saved drawings; remove those drawings and save it again; open it and find that there are no drawings. In the Jet-side regression/integration tests we should also test that Excel can open the workbook at each of these stages.

EPPlusTest/DrawingTest.cs
```diff
@@ -709,6 +710,146 @@ public void DrawingRowheightDynamic()
...
+			Assert.AreEqual(hexagonDrawing, retrievedDrawing);
+			retrievedDrawing = worksheet.Drawings["Picture"];
+			Assert.AreEqual(pictureDrawing, retrievedDrawing);
+			Assert.AreEqual(3, worksheet.Drawings.Count);
+			retrievedDrawing = worksheet.Drawings["NonExistent Drawing"];
+			Assert.AreEqual(null, retrievedDrawing);
+		}
+
+		[TestMethod]
+		public void GetDrawingFromNewWorkbook()
```
emdelaney: I might call this ...FromEmptyWorkbook.

EPPlusTest/DrawingTest.cs
ZacharyPF: agreed

EPPlusTest/DrawingTest.cs
ZacharyPF: Once again, agreed :D

PR #106, opened 2017-09-19
Adds support and tests for cross-sheet sparklines.

EvanSchallerer:
- Fixes corrupt workbooks where sparkline formulas reference another sheet
- Fixes sparkline formula issue when a worksheet is renamed


EPPlus/ExcelRangeBase.cs
```diff
@@ -2092,14 +2092,28 @@ public void Copy(ExcelRangeBase destination, ExcelRangeCopyOptionFlags? excelRan
...
 					}
 				}
-				group.Sparklines.AddRange(newSparklines);
+				if (newSparklines.Count > 0)
+				{
+					if (destination.Worksheet.Equals(this.Worksheet))
+						group.Sparklines.AddRange(newSparklines);
+					else
+					{
+						var newGroup = new ExcelSparklineGroup(destination.Worksheet, group.NameSpaceManager, group.TopNode);
```
emdelaney: Do we need to be cloning the topNode here? If you make a change to the newGroup, is the oldGroup changed as well? This may not be obvious without saving and re-opening the workbook.

EPPlus/ExcelRangeBase.cs
```diff
@@ -2092,14 +2092,28 @@ public void Copy(ExcelRangeBase destination, ExcelRangeCopyOptionFlags? excelRan
...
+						newGroup.Sparklines.Clear();
+						newGroup.Sparklines.AddRange(newSparklines);
+						if (destination.Worksheet.SparklineGroups.TopNode == null)
+							destination.Worksheet.SparklineGroups.TopNode = newGroup.TopNode;
```
emdelaney: This also seems really sketchy to me, we could probably be more intentional about how we generate a topNode for these guys. Maybe copy the inner XML and generate a new node (and add it to the right place in the corresponding worksheet XML) for the topNode? It's really weird / scary to me that this appears to work.

EPPlus/ExcelRangeBase.cs
```diff
@@ -2092,14 +2092,28 @@ public void Copy(ExcelRangeBase destination, ExcelRangeCopyOptionFlags? excelRan
 					{
 						ExcelRangeBase.SplitAddress(sparkline.Formula.Address, out string workbook, out string worksheet, out string address);
 						var newFormula = this.myWorkbook.Package.FormulaManager.UpdateFormulaReferences(address, destination._fromRow - this._fromRow, destination._fromCol - this._fromCol, 0, 0, this.WorkSheet, this.WorkSheet, true);
-						if (!string.IsNullOrEmpty(worksheet) && worksheet.Equals(this.WorkSheet))
```
emdelaney: We're getting rid of this because we know it already collides with the current worksheet? I'm a little confused about why we're only updating the address if the worksheet is fully specified. (Are we just maintaining reference-type consistency / preserving fully-qualified names?).

EPPlus/ExcelRangeBase.cs
emdelaney: We could / should probably add some unit tests where we add a different number of rows to the origin and destination sheets after we copy a worksheet (vaguely approximating a sheet-and-row-replicator case), and ensure that all the sparklines are in the correct spots with the correct sheet names.

EPPlusTest/ExcelWorksheetTests.cs
```diff
@@ -5293,6 +5293,79 @@ public void RenameWorksheetUpdatesExcel2016ChartSeriesAndOtherNamedRanges()
...
+				</x14:sparklineGroup>
+			</sparklineGroups>");
+			var topNode = doc.FirstChild.FirstChild;
+			using (ExcelPackage package = new ExcelPackage())
+			{
+				var worksheet1 = package.Workbook.Worksheets.Add("Sheet1");
+				var worksheet2 = package.Workbook.Worksheets.Add("Sheet2");
+				worksheet1.SparklineGroups.SparklineGroups.Add(new OfficeOpenXml.Drawing.Sparkline.ExcelSparklineGroup(worksheet1, nsManager, topNode));
+				worksheet1.Name = "a new name";
+				Assert.AreEqual(worksheet1.Name, worksheet1.SparklineGroups.SparklineGroups[0].Sparklines[0].HostCell.WorkSheet);
```
emdelaney: We should also add some sparklines to worksheet2 here and test that they do not change :)

EPPlusTest/ExcelRangeBaseTest.cs
```diff
@@ -76,6 +76,104 @@ public void CopyCopiesCommentsFromMultiCellRanges()
...
+
+			var sparklineGroups = worksheet1.SparklineGroups.SparklineGroups;
+			Assert.AreEqual(1, sparklineGroups.Count);
+			var sparkLines = sparklineGroups[0].Sparklines;
+			Assert.AreEqual(2, sparkLines.Count);
+			var originalSparkline = sparkLines[0];
+			var newSparkline = sparkLines[1];
+			Assert.AreEqual(new ExcelAddress(worksheet1.Name, "C19").FullAddress, originalSparkline.HostCell.FullAddress);
+			Assert.AreEqual(new ExcelAddress(worksheet1.Name, "B2").FullAddress, newSparkline.HostCell.FullAddress);
+			Assert.AreEqual(originalSparkline.Formula.WorkSheet, newSparkline.Formula.WorkSheet);
```
emdelaney: Can we test these against worksheet1.Name, or null, or whatever value they're expected to be, rather than against each other?

EPPlusTest/ExcelRangeBaseTest.cs
```diff
@@ -76,6 +76,104 @@ public void CopyCopiesCommentsFromMultiCellRanges()
...
+			Assert.AreEqual(1, originalSparklines.Count);
+			var originalSparkline = originalSparklines[0];
+			Assert.AreEqual(new ExcelAddress(worksheet1.Name, "C19").FullAddress, originalSparkline.HostCell.FullAddress);
+
+			var newSparklineGroups = worksheet2.SparklineGroups.SparklineGroups;
+			Assert.AreEqual(1, newSparklineGroups.Count);
+			var newSparklines = newSparklineGroups[0].Sparklines;
+			Assert.AreEqual(1, newSparklines.Count);
+			var newSparkline = newSparklines[0];
+			Assert.AreEqual(originalSparkline.Formula.WorkSheet, newSparkline.Formula.WorkSheet);
```
emdelaney: Same here, can we test these against more conclusive values like worksheet1.Name instead of the original sparkline's name?

EPPlusTest/ExcelWorksheetTests.cs
```diff
@@ -5293,6 +5293,79 @@ public void RenameWorksheetUpdatesExcel2016ChartSeriesAndOtherNamedRanges()
...
+				Assert.AreEqual(worksheet2.Name, worksheet1.SparklineGroups.SparklineGroups[0].Sparklines[0].Formula.WorkSheet);
+			}
+		}
+
```
emdelaney: We should also throw some test coverage in here where we save the workbook to a temporary file, re-open it, and assert that all the sparklines still exist and have the right values.

EPPlus/ExcelRangeBase.cs
```diff
@@ -2355,12 +2358,10 @@ public void AutoFitColumns(double MinimumWidth, double MaximumWidth)
 		#endregion
 
 		#region IDisposable Members
-
 		public void Dispose()
```
emdelaney: This whole #region can go away.

EvanSchallerer: Will do. Is there any reason to leave the commented out body of Dispose() as well?

EPPlus/Drawing/Sparkline/ExcelSparkline.cs
```diff
@@ -65,28 +65,49 @@ public class ExcelSparkline : XmlHelper
 		/// <param name="topNode">The x14:Sparkline node containing information about the sparkline.</param>
 		public ExcelSparkline(ExcelSparklineGroup group, XmlNamespaceManager nameSpaceManager, XmlNode topNode) : base(nameSpaceManager, topNode)
 		{
-			if (group == null)
```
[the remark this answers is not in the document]
EvanSchallerer: I agree, the ```??? throw``` does seem less intentional and is easier to miss, as well as being less consistent in that it can't be used for arguments if you are doing anything other than assigning them. I do that mostly to get rid of the annoying warnings, so I would be in favor of a config update if that's the convention that you prefer :)

EPPlus/Drawing/Sparkline/ExcelSparklineGroup.cs
```diff
@@ -195,6 +195,74 @@ public void Save()
...
+			attribute = topNode.Attributes["minAxisType"];
+			if (attribute != null)
+				this.MinAxisType = ParseSparklineAxisMinMax(attribute.InnerText);
+			attribute = topNode.Attributes["maxAxisType"];
+			if (attribute != null)
+				this.MaxAxisType = ParseSparklineAxisMinMax(attribute.InnerText);
+			this.RightToLeft = ExcelConditionalFormattingHelper.GetAttributeBool(topNode, "rightToLeft");
+			if (copyChildren)
+			{
+				// Parse the actual sparklines.
```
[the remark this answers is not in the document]
EvanSchallerer: I used it to copy the style of the node while reusing some existing code, but after you pointed that out I think it makes more sense to rename it and move copying the children outside of this method.

EPPlus/Drawing/Sparkline/ExcelSparkline.cs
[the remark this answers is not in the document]
emdelaney: Yeah, let's do that :) The .editorconfig file should be your friend in this endeavor.

PR #107, opened 2017-09-25
Fixed issue where relative named range references were not handled correctly (or at all, really).

maxackley:


EPPlus/ExcelNamedRange.cs
```diff
@@ -130,9 +130,45 @@ public int ActualSheetID
...
+		#endregion
+
+		#region Public Methods
+		/// <summary>
+		/// Gets the address relative to the specified row and column.
+		/// </summary>
+		/// <param name="relativeRow">The row from which the named range is used.</param>
+		/// <param name="relativeColumn">The column from which the named range is used.</param>
+		/// <returns>the address relative to the specified row and column.</returns>
+		public string GetRelativeAddress(int relativeRow, int relativeColumn)
```
emdelaney: Some unit tests for this public method would be great :) We should probably test some cross-sheet cases in addition to the straightforward ones.

PR #110, opened 2017-10-04
Removes conditionalFormattingRule grouping by address, adds test

EvanSchallerer:
Removes functionality from ExcelConditionalFormattingRule that groups rules by address as this can corrupt a workbook.


EPPlus/ConditionalFormatting/Rules/ExcelConditionalFormattingRule.cs
```diff
@@ -168,49 +168,32 @@ public XmlNode Node
 		}
 
 		/// <summary>
-		/// Address of the conditional formatting rule
+		/// Address of the conditional formatting rule.
 		/// </summary>
 		/// <remarks>
```
emdelaney: I haven't noticed this `remarks` section before; we might want to think about utilizing this field more often :)

PR #113, opened 2017-10-10
Updates DependencyChainFactory to add relative named range dependencies

EvanSchallerer:
- Adds dependencies of relative named ranges to dependency chain
- Adds tests for a relative named range with dependencies

### DIFF
--- a/EPPlusTest/ExcelNamedRangeCollectionTest.cs
+++ b/EPPlusTest/ExcelNamedRangeCollectionTest.cs
@@ -868,6 +868,36 @@ namespace EPPlusTest
 				Assert.AreEqual("Correct!", sheet.Cells[4, 4].Value);
 			}
 		}
+
+		[TestMethod]
+		public void RelativeNamedRangeResolvesDependencies()
+		{
+			using (var excelPackage = new ExcelPackage())
+			{
+				var sheet = excelPackage.Workbook.Worksheets.Add("Sheet1");
+				excelPackage.Workbook.Names.Add("RelativeNamedRange", new ExcelRangeBase(sheet, $"$E1"));
+				sheet.Cells[3, 4].Formula = "RelativeNamedRange";
+				sheet.Cells[3, 5].Formula = "F3";
+				sheet.Cells[3, 6].Value = "Correct!";
+				sheet.Calculate();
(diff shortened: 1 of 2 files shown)


EPPlusTest/ExcelNamedRangeCollectionTest.cs
```diff
@@ -868,6 +868,36 @@ public void RelativeNamedRangeWithNegativeRowAndPositiveColumnOffsetResolvesCorr
...
+
+		[TestMethod]
+		public void RelativeNamedRangeResolvesDependencies()
+		{
+			using (var excelPackage = new ExcelPackage())
+			{
+				var sheet = excelPackage.Workbook.Worksheets.Add("Sheet1");
+				excelPackage.Workbook.Names.Add("RelativeNamedRange", new ExcelRangeBase(sheet, $"$E1"));
+				sheet.Cells[3, 4].Formula = "RelativeNamedRange";
+				sheet.Cells[3, 5].Formula = "F3";
```
ZacharyPF: What's the point of this?

EPPlusTest/ExcelNamedRangeCollectionTest.cs
```diff
@@ -868,6 +868,36 @@ public void RelativeNamedRangeWithNegativeRowAndPositiveColumnOffsetResolvesCorr
...
+
+		[TestMethod]
+		public void ReferencedRelativeNamedRangeResolvesDependencies()
+		{
+			using (var excelPackage = new ExcelPackage())
+			{
+				var sheet = excelPackage.Workbook.Worksheets.Add("Sheet1");
+				excelPackage.Workbook.Names.Add("RelativeNamedRange", new ExcelRangeBase(sheet, $"$E1"));
+				sheet.Cells[3, 4].Formula = @"IF(RelativeNamedRange=""Correct!"", true, false)";
+				sheet.Cells[3, 5].Formula = "F3";
```
ZacharyPF: See previous.

EPPlusTest/ExcelNamedRangeCollectionTest.cs
EvanSchallerer: Of the test in general? This failed before the changes because the reference would not be resolved before the named range.

PR #118, opened 2017-12-07
Allow excel addresses to fully-qualify the sheet name of an end refe…

emdelaney:
…rence, such as Sheet1!A1:Sheet1!J10.

### DIFF
--- a/EPPlus/ExcelAddress.cs
+++ b/EPPlus/ExcelAddress.cs
@@ -730,6 +730,7 @@ namespace OfficeOpenXml
 
 		private bool ExtractAddress(string fullAddress)
 		{
+			fullAddress = Regex.Replace(fullAddress, ":[^!,]*!", ":"); // strip sheet names from overly-qualified references like A1:Sheet1!A2 or A4:'sheet name'!B5.
 			var brackPos = new Stack<int>();
 			var bracketParts = new List<string>();
 			string first = "", second = "";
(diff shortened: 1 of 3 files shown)


EPPlus/ExcelAddress.cs
```diff
@@ -730,6 +730,7 @@ private void Validate()
 
 		private bool ExtractAddress(string fullAddress)
 		{
+			fullAddress = Regex.Replace(fullAddress, ":[^!,]*!", ":"); // strip sheet names from overly-qualified references like A1:Sheet1!A2 or A4:'sheet name'!B5.
```
ZacharyPF: Why are we excluding a comma in our search?

emdelaney: I'm excluding commas because you can pass in non-adjacent addresses like A1, B2, C3:D4, Sheet1!D5, Sheet1!D6:Sheet1!D7, etc.

EPPlusTest/ExcelAddressTest.cs
```diff
@@ -8,6 +8,36 @@ public class ExcelAddressTest
 	{
 		#region ExcelAddressBase Tests
 		#region Address Tests
+		[TestMethod]
```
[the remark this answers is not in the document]
emdelaney: No reason not to add it :) I'll do so now.

PR #119, opened 2017-12-11
Add support for deleting external references.

ZacharyPF:


EPPlus/ExternalReferenceCollection.cs
```diff
@@ -0,0 +1,109 @@
...
+		/// Removes an external reference from the workbook by ID.
+		/// </summary>
+		/// <param name="id">The ID of the reference to delete.</param>
+		public void DeleteReference(int id)
+		{
+			var reference = this.PrivateReferences.FirstOrDefault(r => r.Id == id);
+			if (reference != null)
+			{
+				this.PrivateReferences.Remove(reference);
+				this.TopNode.RemoveChild(reference.Node);
```
emdelaney: Do we need to delete any external link parts/documents from xl/externalLinks/* in addition to removing the node from the workbook?

ZacharyPF: Nope. Well, depends on how correct you want to be. I did a bunch of experiments and proved that Excel will clean up all that stuff for you. All you need to do is make sure the workbook.xml reference is gone.

PR #120, opened 2017-12-11
Fixes null reference in data validation.

EvanSchallerer:
Fixes the null reference that can occur when data validation is used in a workbook with multiple worksheets.

### DIFF
--- a/EPPlus/ExcelWorksheet.cs
+++ b/EPPlus/ExcelWorksheet.cs
@@ -4827,13 +4827,12 @@ namespace OfficeOpenXml
 			{
 				for (int i = sheet.DataValidations.Count - 1; i >= 0; i--)
 				{
-					var validation = sheet.DataValidations.ElementAt(i) as DataValidation.Contracts.IExcelDataValidationList;
-					if (validation != null)
+					if (sheet.DataValidations.ElementAt(i) is DataValidation.Contracts.IExcelDataValidationList validation)
 					{
 						string worksheetName = "!";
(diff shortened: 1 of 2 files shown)


EPPlus/ExcelWorksheet.cs
```diff
@@ -4827,13 +4827,12 @@ private void UpdateDataValidationRanges(int rowFrom, int rows, int columnFrom, i
...
 				{
-					var validation = sheet.DataValidations.ElementAt(i) as DataValidation.Contracts.IExcelDataValidationList;
-					if (validation != null)
+					if (sheet.DataValidations.ElementAt(i) is DataValidation.Contracts.IExcelDataValidationList validation)
 					{
 						string worksheetName = "!";
 						if (validation.Address.WorkSheet == null && sheet.Name.ToUpper() == this.Name.ToUpper()) //This formula references the sheet it is on
 							worksheetName = sheet.Name;
-						else if (validation.Address.WorkSheet.ToUpper() == this.Name.ToUpper()) //This formula references another sheet in the workbook
+						else if (validation.Address.WorkSheet != null && validation.Address.WorkSheet.ToUpper() == this.Name.ToUpper()) //This formula references another sheet in the workbook
```
ZacharyPF: you could do:
validation.Address.WorkSheet?.ToUpper() == this.Name.ToUpper()

EvanSchallerer: I wanted to be more explicit in case this.Name.ToUpper() was null, but that probably is safe enough.

PR #122, opened 2017-12-14
Fix logical functions to expand range arguments when evaluating.

EvanSchallerer:
- Expand range arguments in logical functions AND and OR
- Add/organize tests 
- Refactor/format other logical functions

### DIFF
--- a/EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
+++ b/EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
@@ -314,8 +314,8 @@ namespace OfficeOpenXml.FormulaParsing.Excel.Functions
 		/// <returns></returns>
 		protected FunctionArgument GetFirstArgument(FunctionArgument argument)
 		{
-			if (argument.Value is List<FunctionArgument> list)
-				return list.First();
+			if (argument.Value is IEnumerable<FunctionArgument> enumerableArgument)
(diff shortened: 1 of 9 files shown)


EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -307,11 +314,8 @@ protected bool ArgToBool(IEnumerable<FunctionArgument> arguments, int index)
 		/// <returns></returns>
 		protected FunctionArgument GetFirstArgument(FunctionArgument argument)
 		{
-			var list = argument.Value as List<FunctionArgument>;
-			if (list != null)
-			{
+			if (argument.Value is List<FunctionArgument> list)
```
emdelaney: Let's check IEnumerable here instead of List, maybe?

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -323,11 +327,8 @@ protected FunctionArgument GetFirstArgument(FunctionArgument argument)
 		/// <returns></returns>
 		protected object GetFirstArgument(object argument)
 		{
-			var list = argument as List<object>;
-			if (list != null)
-			{
+			if (argument is List<object> list)
```
emdelaney: Same here.

PR #125, opened 2017-12-18
Ensure external workbook references result in #REF! during function e…

ZacharyPF:
…valuation.


EPPlusTest/WorkbookTests.cs
```diff
@@ -39,9 +39,10 @@ public void WorkbookReferenceInNamedRangeDoesNotCorruptWorkbook()
 		}
 
 		[TestMethod]
+		[DeploymentItem(@"..\..\Workbooks\externalreferences.xlsx")]
```
emdelaney: What do you think about adding an integration test where we verify that an external workbook reference sets the cell value to #REF!?

ZacharyPF: Good call, I'll add one in there.

PR #129, opened 2018-01-11
Updated conditional formatting with space seperated addresses to sele…

shawncrabtree:
…ct and work correctly.

Addresses are saved to the xml as space seperated, but selected as comma separated because the addresses are unioned.


EPPlus/ConditionalFormatting/X14CondtionalFormattingRule.cs
```diff
@@ -20,11 +20,12 @@ public string Address
 		{
 			get
 			{
-				return this.GetXmlNodeString("xm:sqref");
+				var sqref = this.GetXmlNodeString("xm:sqref");
+				return sqref.Replace(" ", ",");
```
EvanSchallerer: This could be bad for references that have sheet names with spaces in them, right?

ZacharyPF: agreed

shawncrabtree: Excel doesn't allow for sheet qualified addresses in conditional formatting addresses since they're already scoped to a worksheet.

PR #130, opened 2018-01-18
Fix MIN/MAX functions to work with decimal values, add tests.

EvanSchallerer:
- Refactor MIN and MAX functions to use same logic
- Add support for decimal datatypes in MIN and MAX
- Fix incorrect MIN/MAX argument count checks
- Fix incorrect MIN/MAX cell reference logic
- Add tests

### DIFF
--- a/EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
+++ b/EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
@@ -145,31 +145,7 @@ namespace OfficeOpenXml.FormulaParsing.Excel.Functions
 		/// <returns>Returns true if there are at least the <paramref name="minLength"/> number of arguments present, and returns false if otherwise.</returns>
 		protected bool ArgumentCountIsValid(IEnumerable<FunctionArgument> arguments, int minLength)
 		{
-			if (arguments == null)
-				return false;
-			else
-				return !this.TooFewArgs(arguments, minLength);
-		}
-
-		private bool TooFewArgs(IEnumerable<FunctionArgument> arguments, int minLength)
-		{
-			{
-				var nArgs = 0;
-				if (arguments.Any())
-				{
-					foreach (var arg in arguments)
-					{
-						nArgs++;
-						if (nArgs >= minLength) return false;
-						if (arg.IsExcelRange)
-						{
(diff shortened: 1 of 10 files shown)


EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
```diff
@@ -145,31 +145,7 @@ protected bool ArgumentsAreValid(IEnumerable<FunctionArgument> arguments, int mi
 		/// <returns>Returns true if there are at least the <paramref name="minLength"/> number of arguments present, and returns false if otherwise.</returns>
 		protected bool ArgumentCountIsValid(IEnumerable<FunctionArgument> arguments, int minLength)
```
shawncrabtree: Want to add some tests for this?

EPPlus/FormulaParsing/Excel/Functions/Math/MinMaxBase.cs
```diff
@@ -0,0 +1,102 @@
...
+					{
+						if (value is FunctionArgument argValue && this.TryGetDoubleValue(argValue.Value, out doubleValue))
+							doubleArgs.Add(doubleValue);
+					}
+				}
+			}
+			return doubleArgs;
+		}
+
+		private bool TryGetDoubleValue(object value, out double doubleValue)
```
shawncrabtree: Can we test this?

EPPlus/FormulaParsing/Excel/Functions/ExcelFunction.cs
EvanSchallerer: It's protected and also a one-liner, so probably not.

EPPlus/FormulaParsing/Excel/Functions/Math/MinMaxBase.cs
EvanSchallerer: Private, so not really. It is tested indirectly through the MIN and MAX tests though.

ZacharyPF: That's why test-first is supposed to be so great. Theoretically you never write code that isn't tested because you only write the code required to make the tests pass :p that's only so practical though... Especially when you're catching up like this.

PR #132, opened 2018-01-23
Named range inheritance refactor

EvanSchallerer:
* Removes named range inheritance from ExcelRange
* Adds support for named range formula reference updating
* Adds support for error handling in calc chain evaluation
* Removes unnecessary ExcelAddress class
* Adds corresponding tests


EPPlus/ExcelRange.cs
```diff
@@ -68,20 +68,9 @@ internal ExcelRange(ExcelWorksheet sheet, int fromRow, int fromCol, int toRow, i
 			get
 			{
 				if (myWorksheet.Names.ContainsKey(Address))
```
shawncrabtree: Maybe check if Address is valid address and only look at named ranges if not. Then if a matching named range is found, return the formula if it's a valid address.

EvanSchallerer: Unfortunately the "IsValid" method in epplus isn't stable enough to use for this (doesn't support comma or space separated addresses), so I'll leave it as is for now.

PR #133, opened 2018-01-24
Add support for referencing named range formulas as addresses, tests.

EvanSchallerer:


EPPlus/ExcelNamedRange.cs
```diff
@@ -175,6 +178,59 @@ public void UpdateFormula(int rowFrom, int colFrom, int rows, int cols, ExcelWor
...
+		/// </summary>
+		/// <returns>The formula as an <see cref="ExcelRangeBase"/> if it is an address, null otherwise.</returns>
+		public ExcelRangeBase GetFormulaAsCellRange()
+		{
+			var stringBuilder = new StringBuilder();
+			var tokens = this.Workbook.FormulaParser.Lexer.Tokenize(this.NameFormula).ToList();
+			for (int i = 0; i < tokens.Count; ++i)
+			{
+				var token = tokens[i];
+				if (token.TokenType == TokenType.ExcelAddress ||
```
shawncrabtree: You can make a named range like "=(Report!C3)" so you might need to consider open and close parens.

EPPlus/ExcelNamedRange.cs
```diff
@@ -175,6 +178,59 @@ public void UpdateFormula(int rowFrom, int colFrom, int rows, int cols, ExcelWor
...
+					}
+					else
+						return null;
+				}
+				else
+					return null;
+			}
+			var addressString = stringBuilder.ToString();
+			ExcelRangeBase.SplitAddress(addressString, out string workbook, out string worksheetName, out _);
+			var worksheet = this.Workbook.Worksheets[worksheetName];
```
shawncrabtree: TryGet in case the worksheetName is null or empty or SplitAddress fails?

EPPlus/ExcelNamedRange.cs
```diff
@@ -187,6 +243,44 @@ private int GetRelativeLocation(bool fixedLocation, int current, int relative, i
...
+			else if (token.Value.StartsWith(Indirect.Name, StringComparison.InvariantCultureIgnoreCase))
+				formula += IndirectAddress.Name;
+			else
+				return false;
+
+			for (i = index + 1; i < tokens.Count; ++i)
+			{
+				token = tokens[i];
+				formula += token.Value;
+				if (token.TokenType == TokenType.OpeningParenthesis)
```
shawncrabtree: What if you have an offset function nested in an offset function?

EvanSchallerer: Looks like that will work as long as a nested offset isn't used as the first argument to an offset due to the way the dependency chain is handling offsets.

EPPlus/FormulaParsing/Excel/Functions/RefAndLookup/IndirectAddress.cs
```diff
@@ -0,0 +1,65 @@
...
+		/// Executes the function with the specified <paramref name="arguments"/> in the specified <paramref name="context"/>.
+		/// </summary>
+		/// <param name="arguments">The arguments to evaluate the function with.</param>
+		/// <param name="context">The context with which to evaluate the function in.</param>
+		/// <returns>A <see cref="CompileResult"/> containing the result of evaluation.</returns>
+		public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, ParsingContext context)
+		{
+			if (this.ArgumentsAreValid(arguments, 1, out eErrorType argumentError) == false)
+				return new CompileResult(argumentError);
+			var addressArgument = ArgToString(arguments, 0);
```
shawncrabtree: base.ArgToString

EPPlus/ExcelNamedRange.cs
EvanSchallerer: This indexer is safe, so it should be fine.

EPPlus/ExcelNamedRange.cs
EvanSchallerer: Added a check for this, the calculation stack doesn't handle this very gracefully for some cases anyways (ex: (Sheet1!B5):Sheet1!C8), so I'm not super concerned about it.

PR #136, opened 2018-02-08
Fixed issue where getting relative formula of a named range throw an …

shawncrabtree:
Fixed issue where getting relative formula of a named range throw an exception if the named range referred to an entire row or column.

### DIFF
--- a/EPPlus/ExcelCellBase.cs
+++ b/EPPlus/ExcelCellBase.cs
@@ -374,7 +374,7 @@ namespace OfficeOpenXml
 		internal static bool GetRowColFromAddress(string CellAddress, out int FromRow, out int FromColumn, out int ToRow, out int ToColumn)
 		{
 			bool fixedFromRow, fixedFromColumn, fixedToRow, fixedToColumn;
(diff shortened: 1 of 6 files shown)


EPPlus/ExcelCellBase.cs
```diff
@@ -425,7 +427,16 @@ internal static bool GetRowColFromAddress(string CellAddress, out int FromRow, o
 				string[] cells = CellAddress.Split(':');
 				ret = GetRowColFromAddress(cells[0], out FromRow, out FromColumn, out fixedFromRow, out fixedFromColumn);
 				if (ret)
+				{
 					ret = GetRowColFromAddress(cells[1], out ToRow, out ToColumn, out fixedToRow, out fixedToColumn);
+					if (ret)
+					{
+						if (FromRow == 0 && ToRow == 0)
```
zpfaltersack: Do these cases need to also capture something like 1 -> MaxRow, 1 -> MaxColumn?

shawncrabtree: No that logic is done below (lines 434 to 441). Also Excel resolves C1:C{MaxRows} to C:C.

PR #139, opened 2018-02-09
Add extended support for named ranges.

EvanSchallerer:
Re-write of named ranges to match excel's behavior.


EPPlus/ExcelNamedRange.cs
```diff
@@ -90,73 +77,178 @@ public int ActualSheetID
...
+		public ExcelWorksheet LocalSheet { get; private set; }
 
 		/// <summary>
-		/// Gets or sets the index value of this named range in its parent <see cref="ExcelNamedRangeCollection"/>.
+		/// Gets the <see cref="ExcelWorkbook"/> that contains this named range.
 		/// </summary>
-		internal int Index { get; set; }
+		internal ExcelWorkbook Workbook { get; private set; }
 
+		// TODO: Is Index needed anymore? Dependency chain IDs for named ranges may not be correct
```
zpfaltersack: Do we know more about this yet?

EPPlus/ExcelNamedRange.cs
```diff
@@ -90,73 +77,178 @@ public int ActualSheetID
...
+			this.Workbook = workbook;
 			this.LocalSheet = nameSheet;
+			this.NameFormula = formula;
 			this.Index = index;
 		}
+		#endregion
+
+		#region Public Methods
+		/// <summary>
+		/// Gets the formula of a named range relative to the specified <paramref name="relativeRow"/> and <paramref name="relativeColumn"/>.
```
zpfaltersack: Gets the tokenized formula...

EPPlus/ExcelNamedRange.cs
```diff
@@ -90,73 +77,178 @@ public int ActualSheetID
...
+			for (int i = 0; i < tokens.Count; ++i)
+			{
+				var token = tokens[i];
+				if (token.TokenType == TokenType.ExcelAddress ||
+					token.TokenType == TokenType.InvalidReference ||
+					token.TokenType == TokenType.Comma)
+				{
+					stringBuilder.Append(token.Value);
+				}
+				else if (token.TokenType == TokenType.OpeningParenthesis || token.TokenType == TokenType.ClosingParenthesis)
```
shawncrabtree: Should these be added to the string builder? Are they important for like combinations of unions and intersections in an address?

EPPlus/ExcelRangeBase.cs
```diff
@@ -246,6 +249,11 @@ private void SetValueAddress(ExcelAddress address, _setValue valueMethod, object
 		#endregion
 
 		#region Properties
+		public static implicit operator string(ExcelRangeBase excelRangeBase)
```
zpfaltersack: comment

EPPlus/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace OfficeOpenXml.Extensions
+{
+	public static class StringExtensions
```
zpfaltersack: comment

EPPlus/ExcelNamedRange.cs
EvanSchallerer: We'll probably ignore that case for now as EPPlus wouldn't be able to handle those cases anyways.

PR #141, opened 2018-02-12
Fixed issue where the ZCellStore was not shifting well when empty pag…

zpfaltersack:
…es existed in the store.

### DIFF
--- a/EPPlus/ZCellStore.cs
+++ b/EPPlus/ZCellStore.cs
@@ -816,12 +816,12 @@ namespace OfficeOpenXml
 				if (amount > 0)
 				{
 					this.DeConstructIndex(this.MaximumUsedIndex, out int sourcePageIndex, out int sourceInnerIndex);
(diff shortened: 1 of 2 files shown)


EPPlus/ZCellStore.cs
```diff
@@ -820,34 +820,30 @@ public void ShiftItems(int index, int amount)
 
 					for (int i = this.MaximumUsedIndex; i >= Math.Max(index, this.MinimumUsedIndex); --i)
 					{
-						var sourcePage = this.Pages[sourcePageIndex];
-						if (sourcePage == null)
+						var targetPageIndex = sourcePageIndex + numberOfPagesToShit;
+						var targetPageInnerIndex = sourceInnerIndex + numberOfInnerIndexToShit;
```
EvanSchallerer: ToShift?

EvanSchallerer: And the one above it :)

PR #149, opened 2018-02-15
Exposed an AllValues method on IRangeInfo that gets all cell values, …

shawncrabtree:
…including null and empty cells.

### DIFF
--- a/EPPlus/FormulaParsing/EpplusExcelDataProvider.cs
+++ b/EPPlus/FormulaParsing/EpplusExcelDataProvider.cs
@@ -179,7 +179,7 @@ namespace OfficeOpenXml.FormulaParsing
 			public IEnumerable<object> AllValues()
 			{
 				var values = new List<object>();
-				if (this.Address.Addresses != null && this.Address.Addresses.Any())
+				if (this.Address.Addresses?.Any() == true)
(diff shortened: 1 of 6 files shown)


EPPlus/FormulaParsing/EpplusExcelDataProvider.cs
```diff
@@ -168,6 +168,42 @@ public object GetOffset(int rowOffset, int colOffset)
...
+			/// Return a list containing the value of each and every cell in this <see cref="RangeInfo"/>.
+			/// This function exists because iterating normally over a <see cref="ExcelDataProvider.IRangeInfo"/>
+			/// and looking at each <see cref="ExcelDataProvider.ICellInfo"/> will not include cells that have not been set
+			/// (ie: cells that have been empty since the workbook's creation). This function works around that issue.
+			/// </summary>
+			/// <returns>The value of every cell in this <see cref="RangeInfo"/>.</returns>
+			public IEnumerable<object> AllValues()
+			{
+				var values = new List<object>();
+				if (this.Address.Addresses != null && this.Address.Addresses.Any())
```
zpfaltersack: this.Address.Addresses?.Any() == true

PR #170, opened 2018-03-09
Bugs/formula set style

EvanSchallerer:
Add option to set cell style based on a range's calculation result.


EPPlus/FormulaParsing/FormulaParser.cs
```diff
@@ -106,50 +104,46 @@ internal virtual object Parse(IEnumerable<Token> tokens, string worksheet, strin
...
-				}
 				try
 				{
 					var compileResult = _compiler.Compile(graph.Expressions);
 					// quick solution for the fact that an excelrange can be returned.
 					var rangeInfo = compileResult.Result as ExcelDataProvider.IRangeInfo;
 					if (rangeInfo == null)
 					{
+						if (compileResult.Result != null)
+							dataType = compileResult.DataType;
```
shawncrabtree: Why not just set this outside of this scope and remove the special casing below?

PR #178, opened 2018-06-27
Update mathematical operator type coercion.

EvanSchallerer:
Update math operators to have a better understanding of Date types.

### DIFF
--- a/EPPlus/FormulaParsing/Excel/Operators/Operator.cs
+++ b/EPPlus/FormulaParsing/Excel/Operators/Operator.cs
@@ -75,8 +75,11 @@ namespace OfficeOpenXml.FormulaParsing.Excel.Operators
 		{
 			get
 			{
-				CompileResult add(CompileResult l, CompileResult r) => 
-					Operator.CalculateNumericalOperator(l, r, (ld, rd, dataType) => new CompileResult(ld + rd, dataType));
+				CompileResult add(CompileResult l, CompileResult r) =>
+					Operator.CalculateNumericalOperator(l, r, () => {
+						var dataType = Operator.ParseAdditiveOperatorDataType(l.DataType, r.DataType);
+						return new CompileResult(l.ResultNumeric + r.ResultNumeric, dataType);
+					});
 				return myPlus ?? (myPlus = new Operator(OperatorType.Plus, Operator.PrecedenceAddSubtract, add));
 			}
 		}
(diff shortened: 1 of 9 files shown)


EPPlus/FormulaParsing/Excel/Operators/Operator.cs
```diff
@@ -424,6 +422,23 @@ private static bool EitherIsError(CompileResult l, CompileResult r, out ExcelErr
...
+		private static DataType ParseGeneralOperatorDataType(DataType leftType, DataType rightType)
+		{
+			return (leftType == DataType.Integer && rightType == DataType.Integer) ? DataType.Integer : DataType.Decimal;
+		}
+
+		private static DataType ParseAdditiveOperatorDataType(DataType leftType, DataType rightType)
+		{
+			DataType dataType;
+			if (leftType == DataType.Date || rightType == DataType.Date)
+				dataType = DataType.Date;
```
maxackley: You could just return in each condition and omit the variable and last else.

EPPlus/FormulaParsing/Excel/Operators/Operator.cs
```diff
@@ -75,8 +75,11 @@ public static IOperator Plus
 		{
 			get
 			{
-				CompileResult add(CompileResult l, CompileResult r) => 
-					Operator.CalculateNumericalOperator(l, r, (ld, rd, dataType) => new CompileResult(ld + rd, dataType));
+				CompileResult add(CompileResult l, CompileResult r) =>
+					Operator.CalculateNumericalOperator(l, r, () => {
```
maxackley: Since this is split into multiple lines, I would still add curly braces even though it's technically a single statement.

EPPlus/FormulaParsing/Excel/Operators/Operator.cs
```diff
@@ -75,8 +75,11 @@ public static IOperator Plus
 		{
 			get
 			{
-				CompileResult add(CompileResult l, CompileResult r) => 
-					Operator.CalculateNumericalOperator(l, r, (ld, rd, dataType) => new CompileResult(ld + rd, dataType));
+				CompileResult add(CompileResult l, CompileResult r) =>
+					Operator.CalculateNumericalOperator(l, r, () => {
+						var dataType = Operator.ParseAdditiveOperatorDataType(l.DataType, r.DataType);
```
maxackley: I would also move this outside of the inner lambda and that way you can inline the operator delegate.

EPPlus/FormulaParsing/Excel/Operators/Operator.cs
maxackley: This would apply to all the operators of course.

PR #182, opened 2018-07-26
Handling operations involving time data types.

shawncrabtree:


EPPlus/FormulaParsing/Excel/Operators/Operator.cs
```diff
@@ -341,7 +342,7 @@ private static CompileResult CalculateNumericalOperator(CompileResult left, Comp
 
 		private static bool IsNumericType(CompileResult result)
 		{
-			return result.DataType == DataType.Integer || result.IsNumeric || result.IsDateString 
+			return result.IsNumeric || result.IsDateString
```
EvanSchallerer: Does result.IsNumeric include integers by default or something?

shawncrabtree: Yes. Deleted that check because it's redundant.

PR #192, opened 2018-10-01
Better memory management with excel worksheet cell data writing

zpfaltersack:
Cleaned up some memory usage with writing worksheet row data back out as XML.
Added a call to the IFormulaManager interface that allows a user to flush any cached data associated with the formula manager instance.


EPPlus/FormulaManager.cs
```diff
@@ -157,6 +168,11 @@ public string UpdateFormulaDeletedSheetReferences(string originalFormula, string
 				throw new ArgumentNullException(nameof(deletedSheetName));
 			return this.UpdateFormulaWorksheetReferences(originalFormula, deletedSheetName, null);
 		}
+
+		public void FlushCache()
```
EvanSchallerer: Comment for this?

shawncrabtree:
💯

PR #201, opened 2018-11-12
Features/tasks/row items clean up

laum18:
Abstracted BuildRowItems method and added more test.


EPPlus/Table/PivotTable/DataCalculation/ColumnGrandTotalHelper.cs
```diff
@@ -0,0 +1,118 @@
...
+		/// Calculate the grand total values when there are multiple row data fields and no column fields. Only used to calculate row grand total values.
+		/// </summary>
+		/// <param name="innerHeaderIndex">The index for the inner 'for' loop.</param>
+		/// <param name="value">The value in the current cell.</param>
+		/// <param name="outerHeader">The <see cref="PivotTableHeader"/>.</param>
+		/// <param name="grandTotal">The current grand total value that needs to be updated.</param>
+		/// <returns>The updated grand total value.</returns>
+		protected override double? CalculateTotalWithNoColumns(int innerHeaderIndex, double value, PivotTableHeader outerHeader, double? grandTotal)
+		{
+			throw new NotImplementedException();
```
EvanSchallerer: Was this intended to be left here?

laum18: Yes, it's a method only used for calculating row grand totals.

EvanSchallerer: That makes sense, it might be worth a comment then.

PR #203, opened 2018-11-26
Fixed bug so that a given index in the SEARCH function correctly returns the searched text.

laum18:


EPPlus/FormulaParsing/Excel/Functions/Text/Search.cs
```diff
@@ -40,7 +40,9 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
 			var startIndex = 0;
 			if (functionArguments.Count() > 2)
 				startIndex = base.ArgToInt(functionArguments, 2);
-			var result = searchIn.IndexOf(search, startIndex, System.StringComparison.OrdinalIgnoreCase);
+			// Subtract 1 because Excel uses 1-based index
+			int index = functionArguments.Count() > 2 ? startIndex - 1 : startIndex;
```
EvanSchallerer: Can you -1 in the if statement and use startIndex instead?

PR #216, opened 2018-12-27
Fix INDEX function to support single-cell references for the array arg.

EvanSchallerer:


EPPlus/FormulaParsing/Excel/Functions/RefAndLookup/Index.cs
```diff
@@ -63,7 +61,11 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
 			}
 			if (rowDataCandidate != null && rowDataCandidate.Value is ExcelErrorValue && rowDataCandidate.Value.ToString() == ExcelErrorValue.Values.NA)
 				return base.CreateResult(rowDataCandidate.Value, DataType.Integer);
-			if (cellRange.IsExcelRange)
+
+			// A single cell array ignores the row number and column number arguments, returning the single value in the array.
+			if (!cellRange.IsExcelRange)
+				return new CompileResult(cellRange.Value, cellRange.DataType);
+			else
```
shawncrabtree: At the end of this method there is a "throw new NotImplementedException()" that can be deleted because it's unreachable code I think.

EvanSchallerer: Good catch.

PR #224, opened 2019-01-16
Features/9963 calculated fields

EvanSchallerer:
Add support for calculated fields when refreshing a pivot table.


EPPlus/ExcelNamedRange.cs
```diff
@@ -46,6 +46,17 @@ public sealed class ExcelNamedRange
 		private string myFormula;
 		#endregion
 
+		#region Constants
+		/// <summary>
+		/// An (mostly) complete list of ASCII characters that are not allowed in Excel named ranges.
+		/// </summary>
+		public static readonly char[] IllegalCharacters =
```
shawncrabtree: Wouldn't it be safer to just have a list of legal characters?

PR #233, opened 2019-01-31
Fix bug for pivot tables that have multiple row fields and only dataf…

EvanSchallerer:
…ield column fields. Add test.


EPPlusTest/Table/PivotTable/ExcelPivotTableTest.cs
```diff
@@ -6134,6 +6220,16 @@ public void PivotTableRefreshCustomSortingWithReferencesToDataFieldsTwoRowFields
 		}
 		#endregion
 
+		[TestMethod]
+		public void SKDjfsdfsdf()
```
zpfaltersack: Excellent test name. Much kudos.

EvanSchallerer: oops...

PR #238, opened 2019-02-15
Tabular form pivot table field settings

laum18:
Completed tabular form field settings for pivot tables.

Note: The methods surrounding tabular form are quite long because there a lot of different cases, but it is straight-forward and easy to read.


EPPlus/Table/PivotTable/ExcelPivotTableField.cs
```diff
@@ -297,10 +297,11 @@ public bool Compact
 
 		/// <summary>
 		/// Gets or sets whether the items in this field should be shown in Outline form.
+		/// A value of false corresponds to "tabular form".
 		/// </summary>
 		public bool Outline
 		{
-			get { return base.GetXmlNodeBool("@outline"); }
+			get { return base.GetXmlNodeBool("@outline", true); }
 			set { base.SetXmlNodeBool("@outline", value); }
```
EvanSchallerer: Set the remove if value in setxmlnodebool as well

PR #247, opened 2019-02-25
Fix bug where Excel stores zero pivot table indentation as 127.

EvanSchallerer:


EPPlus/Table/PivotTable/ExcelPivotTable.cs
```diff
@@ -1786,7 +1786,12 @@ private void WriteRowHeaders(StringResources stringResources)
 				this.Worksheet.Cells[dataRow++, this.Address.Start.Column].Value = this.DataFields.First().Name;
 		}
 
-		private int GetIndent(int depth) => depth * (this.Indent + 1);
+		private int GetIndent(int depth)
+		{
+			// Excel stores a zero indent as 127, and other values are stored as 1 less than the UI shows.
+			var indent = this.Indent == 127 ? 0 : this.Indent + 1;
```
shawncrabtree: I bet you the workbook corrupts if the indent is more than 127. Probably not something to worry about though.

EvanSchallerer: Since we aren't setting the value anywhere other than tests I'll leave it as is. Good catch though.

PR #261, opened 2019-03-13
Blank values in pivot tables matches Excel.

laum18:
Blank rows in the source data table shows up in the pivot table as "(blank)".


EPPlus/Table/PivotTable/ExcelPivotTable.cs
```diff
@@ -1487,6 +1487,7 @@ private string GetItemValueByGroupingType(string sharedItemValue, PivotFieldDate
 			Dictionary<int, List<int>> cacheRecordPageFieldIndices, CacheRecordNode cacheRecord, string searchValue, CacheFieldNode cacheField = null)
 		{
 			// If an identical child already exists, continue. Otherwise, create a new child.
+			searchValue = searchValue ?? "(blank)";
```
zpfaltersack: Does this need to be localized?

laum18: Yes it does.

PR #279, opened 2019-04-02
Fix issues related to pivot table cached value double precision.

EvanSchallerer:


EPPlus/Table/PivotTable/CacheFieldNode.cs
```diff
@@ -36,6 +36,11 @@ namespace OfficeOpenXml.Table.PivotTable
 	/// </summary>
 	public class CacheFieldNode : XmlHelper
 	{
+		#region Constants
+		// Excel stores numbers to 10^-15 precision. Storing 10^-14 to avoid any chance of corruption.
+		private const double Epsilon = 0.00000000000001;
```
zpfaltersack: You missed a 0

PR #285, opened 2019-04-05
Fix pivot table calculated fields that reference dates.

EvanSchallerer:


EPPlus/Table/PivotTable/DataCalculation/TotalFunctionHelper.cs
```diff
@@ -166,14 +166,25 @@ public object EvaluateCalculatedFieldFormula(Dictionary<string, List<object>> na
...
 			{
 				var excelName = nameToValues.Key;
 				if (this.FieldNamesToSanitizedFieldNames.ContainsKey(excelName))
 					excelName = this.FieldNamesToSanitizedFieldNames[excelName];
-				// Update the formula of the named range with the name of the field 
-				// to be the sum of the field values.
-				this.TempWorksheet.Names[excelName].NameFormula = $"SUM({string.Join(",", nameToValues.Value)})";
+
+				// Convert DateTime values to OADates in order for calculations to proceed correctly.
+				stringValues.Clear();
```
shawncrabtree: Just declare stringValues in this scope instead of the parent scope. Then you don't have to clear.

PR #315, opened 2019-07-30
Add support and tests for currency-formatted string VALUE function arguments.

EvanSchallerer:


EPPlus/FormulaParsing/Excel/Functions/Text/Value.cs
```diff
@@ -63,7 +63,7 @@ public override CompileResult Execute(IEnumerable<FunctionArgument> arguments, P
 					result = double.Parse(value);
 				return CreateResult(result, DataType.Decimal);
 			}
-			if (double.TryParse(value, NumberStyles.Float, CultureInfo.CurrentCulture, out result))
+			if (double.TryParse(value, NumberStyles.Float | NumberStyles.Currency, CultureInfo.CurrentCulture, out result))
 				return CreateResult(result, DataType.Decimal);
```
maxackley: Do we want to fully qualify the calls to CreateResult in this method? I'm not sure if that convention applies to the EPPlus project.

